PR #31, opened 2014-12-11
Use common English salutations

alasdaircr:
Ms. and Miss are common british salutations, Drs. is not so common

### DIFF
--- a/application/Espo/Modules/Crm/Resources/metadata/entityDefs/Contact.json
+++ b/application/Espo/Modules/Crm/Resources/metadata/entityDefs/Contact.json
@@ -5,7 +5,7 @@
         },
         "salutationName": {
             "type": "enum",
-            "options": ["", "Mr.", "Mrs.", "Dr.", "Drs."]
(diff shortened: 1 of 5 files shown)


alasdaircr:
Were you not interested in this?

yurikuzn:
The existing list is enough we think. Users can manage the list in Field Manager.

alcoheca:
Well in terms of the English language, Mr, Ms, Miss and Mrs are all primary and should be there. The rest is just superfluous.

Drs. doesn't exist in English (if you mean female doctor)

yurikuzn:
You're right about Drs. I think we need to add Prof. but Miss is unnecessary.

yurikuzn:
https://github.com/espocrm/espocrm/commit/1c7785f6ba5bd9e5c1e939b36280c17a0e3e3b3d

alasdaircr:
Some women still use Miss instead of Ms, but this is a step in the right direction. thanks

PR #32, opened 2015-01-06
Allow filtering array field types. fixes #17

### LINKED ISSUE #17: Cannot Add Custom Field of Type Array to Search Filter
reported here: http://forum.espocrm.com/bug-reports/2913-cannot-add-custom-field-of-type-array-to-search-filter?routestring=forum/bug-reports/2913-cannot-add-custom-field-of-type-array-to-search-filter

Leaving this as placeholder as I need this and so will make up a PR

alasdaircr:


alasdaircr:
One thing which needs fixed is the multi select UI element is not being re-selected when you navigate away and then back. The actual results reflect the search parameters but the UI doesn't show this.

PR #33, opened 2015-01-16
Send emails when mentioned, can be toggled in user preferences

alasdaircr:
This could be improved by sharing more code with the EmailAssignement job but I didn't want to refactor too much.

### DIFF
--- a/application/Espo/Hooks/Note/Mentions.php
+++ b/application/Espo/Hooks/Note/Mentions.php
@@ -106,6 +106,19 @@ class Mentions extends \Espo\Core\Hooks\Base
     protected function notifyAboutMention(Entity $entity, \Espo\Entities\User $user)
     {
         $this->getNotificationService()->notifyAboutMentionInPost($user->id, $entity->id);
+        $job = $this->getEntityManager()->getEntity('Job');
+        $job->set(array(
+            'serviceName' => 'EmailNotification',
+            'method' => 'notifyAboutMentionJob',
+            'data' => json_encode(array(
+                'userId' => $user->id,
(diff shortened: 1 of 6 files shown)


alasdaircr:
Maybe long term, it would be good to check if the user is *active* on Espo at the moment the mention is made. Not sure how tricky that would be

yurikuzn:
I planned to implement unified email notifications for all types of notifications.

alasdaircr:
Amazing that would be even better!

yurikuzn:
I thought about what if to send notifications grouped into a single email message. To avoid sending a lot of emails to the user.

alasdaircr:
That would be great. With some polishing, could make a daily digest email from this sort of code.

yurikuzn:
I didn't mean sort of daily digest, but it''s a good idea too.
I meant sending actual notifications grouped each minute. To avoid sending few emails like:
1. User 'Tester' create entity 'Foo'.
2. User 'Tester' assigned entity 'Foo' to you.
3. User 'Tester' changed status of entity 'Foo' to 'Assigned'

yurikuzn:
And also it would be good not to send email notification if it was already read in application.

alasdaircr:
Definitely, that's a good idea.

I thought of note_viewed link table so that when an event/note is viewed in app then no email is sent.

In terms of emails the trick is how to not spam too much, but not too take too long to send an email. For instance even waiting a minute for an email to tell you that a task has been assigned is too much for some people - they think the process has failed somehow.

So maybe only buffer 20-30 seconds of activity or so?

yurikuzn:
Bad that cron can't be run more often than once a minute.

PR #38, opened 2015-03-13
INSERTing number fix for PhoneNumber.php

Threxxy:
Adding a new phone number to a contact was giving the error: SQLSTATE[HY000]: General error: 1366 Incorrect integer value: '' for column 'primary' at row 1

Cast the boolean to an int and it appears to be fixed.


yurikuzn:
Hi.
Never had this problem. What version of MySQL and php do you have?

Threxxy:
Thanks for merge.
PHP is: PHP 5.5.9
MySQL is:  5.7.6

yurikuzn:
Thank you

PR #40, opened 2015-04-09
Dashlet - my delegated tasks

alasdaircr:
hi suggesting this as a dashlet, but it needs a fix.

This is an old patch I've been using for months, but am now rebasing everything against master (the tab->spaces change stopped me doing this sooner).

This dashlet is supposed to show you tasks that are created by you, but not assigned to you. The second part is not working yet as the where clause syntax has changed.

### DIFF
--- a/application/Espo/Core/SelectManagers/Base.php
+++ b/application/Espo/Core/SelectManagers/Base.php
@@ -578,5 +578,15 @@ class Base
             'assignedUserId' => $this->getUser()->id
         );
     }
+
+    protected function getBoolFilterWhereOnlyMyCreated()
+    {
+        return array(
+            'type' => 'equals',
+            'field' => 'createdById',
+            'value' => $this->user->id,
+        );
+    }
+
 }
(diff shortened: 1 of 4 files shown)


yurikuzn:
Hi

Not sure whether we need it built-in. Will be good to have more enhanced Tasks and Activities dashlets in future. Where user can switch between modes. Not sure. What do you think? Or keeping them simple is a better choice?

alasdaircr:
Hi Yuri

I think improved dashlets would be great, perhaps getting a little bit closer to what is available with the reports extension?

In the meantime though, do you have a suggestion on how to modify the where clause so that it is ```created_by == ME && assigned_to != ME``` ?

yurikuzn:
You can create SelectManager for Task in Custom directory with your filter.
```
$result['whereClause'][] = array(
  'createdById=' => $this->getUser()->id,
  'assignedUseId!=' => $this->getUser()->id,
);
````

PR #47, opened 2015-04-28
Relation between Emails and Contacts

feldi88:
There were two small bugs in the relation between emails and contacts:
* in the contact entity, the relation to email was wrong due to a copy & paste misstake
* in the email entity, contact was not allowed as parent entity


yurikuzn:
Hi

First is really bug. The second is not. It's designed so.

Thanks.

yurikuzn:
It's supposed that in B2C model administrator should add Contact to parent option lists. And remove Account entity from everywhere. Also there is a bollean parameter `b2cMode` in config.

PR #68, opened 2015-11-12
FIX: Couldn't add unique index due to prior duplicates

alasdaircr:
In the slim chance that duplicates already existed, it's impossible to rebuild an old database.

(Happened to me earlier this year)


tmachyshyn:
Perhaps the error occurred because of your some custom code. I didn't see this error from anyone else.

PR #155, opened 2016-07-08
Fix utf-8 csv export to excel

olexus:


grzegorzgie:
The fact is not working properly. Also I have a problem in Polish.

yurikuzn:
Is it not working with this fix applied?

grzegorzgie:
Unfortunately, the latest version , which checked the export did not work properly . When was the amendment ?

yurikuzn:
No enough information about the issue. Tabulation lacks of one space.

PR #157, opened 2016-07-13
Move hard coded ldap config to config.php

hi-ko:
Tested with MS AD and samba4
s. 
http://forum.espocrm.com/forum/installation-and-upgrade-help/4567-espo-and-ldap-sso
http://forum.espocrm.com/forum/installation-and-upgrade-help/5231-mapping-ldap-attributes-for-new-users

TODO: At least ldapUserNameAttribute config attribute should be added to Admin/authentication to allow admin to change the ldapUserNameAttribute from the user interface instead of changing the config.php


hi-ko:
fix: sorry - the config load block has to be moved after constructor.

PR #177, opened 2016-08-03
Enhance the header to be more responcive

ayman-alkom:
I tried with this and found it very useful, I think we dont have to use the bootstrap grid in this part

image


yurikuzn:
Unfortunately there is an issue with margins width and with long record names. We will consider to improve header in future.

PR #181, opened 2016-08-06
add navbar height parameter

ayman-alkom:
This was very important in the case we change the theme navbar height from CSS

Every theme now has its own bar height

I also deleted the interval function `processUpdateWidth()`, because it reduce the performance and the resize evet is enough


yurikuzn:
We can't remove processUpdateWidth here.

yurikuzn:
I will add navbarHeight. Thank for suggestion. We missed this point before.

PR #221, opened 2016-09-02
Add CONCAT function to the base query class

ayman-alkom:
just an idea, then we can using it like this:
```php
        $selectParams = array(
            'select' => [
                .........................
                ['VALUE:', 'dateEndDate'],
                ['CONCAT:users.first_name, " ", users.last_name','assignedUserName'],
                ................
            ],
            'leftJoins' => ['users'],
            'whereClause' => .................
```


yurikuzn:
Missing quoting. Possible SQL injection. But idea is good. Thanks.

yurikuzn:
Will be added in 5.6.0

ayman-alkom:
thanks a lot

PR #240, opened 2016-09-20
afterRemove hook didn't remove all notes, leading to 404s and blank e…

alasdaircr:
…ntity names

### DIFF
--- a/application/Espo/Hooks/Common/Stream.php
+++ b/application/Espo/Hooks/Common/Stream.php
@@ -82,9 +82,13 @@ class Stream extends \Espo\Core\Hooks\Base
             $this->getStreamService()->unfollowAllUsersFromEntity($entity);
         }
         $query = $this->getEntityManager()->getQuery();
+        $quotedId = $query->quote($entity->id);
+        $quotedType = $query->quote($entity->getEntityType());
         $sql = "
             DELETE FROM `note`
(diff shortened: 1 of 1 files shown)


yurikuzn:
I think we need to opt to setting deleted = 1 instead of removing here.

alasdaircr:
Sure - then eventually could be undo'd which would be great. I'll change this soon

yurikuzn:
I'm concerning about next. Note related to opportunity through parent and related to account through superParent. If we remove account we will lost notes of opportunity.

yurikuzn:
I've just remembered that that was a reason why I didn't make deleting by super parent.

alasdaircr:
I see. A MUCH better option :-) would be to handle the deleted entities and show them in the notes, but as strikethrough.

And a note should be left for deletions with the same strikethrough effect on the entity name.

Also, any deleted entity should not be clickable as the 404 page is not that helpful for people who don't understand what a 404 is...

alasdaircr:
e.g.

**John** mentioned you in relation to **Contact** ~~Brian Gone~~

alasdaircr:
closing this then as it turned into more of a feature request - fixing notes to deleted entites

yurikuzn:
https://github.com/espocrm/espocrm/issues/245

PR #250, opened 2016-09-29
Rename Number Class to NumberUtil for not scalar type with PHP7 Number Class

ayman-alkom:


yurikuzn:
We will fix it separately. Thanks.

PR #255, opened 2016-10-03
safe call of function

ayman-alkom:


yurikuzn:
Why is it safer? I don't see the point.

ayman-alkom:
In fact I am too, it is just something related to syntax maybe.
feel free to close it if not liked it :)

yurikuzn:
It was actual on previous versions of PHP. Now it's not.

PR #269, opened 2016-10-12
Fix de_DE translation

kaechele:
There also was a bug in po.js that would result in a drop of any translation string that legitimately contained a backslash (\\) as soon as the lang.js step is run.


po.js
```diff
@@ -175,6 +175,7 @@ PO.prototype.replaceAll = function (string, find, replace) {
 }
 
 PO.prototype.fixString = function (savedString) {
+    savedString = this.replaceAll(savedString, "\\", '\\\\');
```
kaechele: There is currently one string that legitimately contains a backslash (\). If the above line is not present the translated string will just never make it into the translation and thus will be presented as untranslated whenever you generate the PO file again.

yurikuzn: Ok. Thank you. Would be good to have such things separately in different pull requests it next time.

yurikuzn:
Thank you for contributing.

PR #295, opened 2016-11-10
PATCH: records not being exported in default UI order

alasdaircr:


yurikuzn:
I've fixed it here: https://github.com/espocrm/espocrm/commit/8a8cb08b998c1aa1fac7f0e3d9a7b84fea6cf92b

Thank you.

PR #309, opened 2016-11-18
Use 'contains' as default text & varchar filter option

alasdaircr:
Contains is the broadest search option, so should really be the default from a user perspective.


yurikuzn:
Reasonable for text fields. Not sure about varchar. 

https://github.com/espocrm/espocrm/commit/afdd15ac445b9df6a2d7919e7c7b79c88573c7c8

donavynelliott:
If not a default, this should at least be toggle-able or editable in some sense or at least have a special case.

PR #310, opened 2016-11-18
PATCH: More date filter options

alasdaircr:
'Before X days', though clumsily written means "At least X days in the past"

likewise with 'After X Days' = "At least X days into the future"

This field may need a refactor as there's other possible options still not covered and the list is huge.

### DIFF
--- a/application/Espo/Core/SelectManagers/Base.php
+++ b/application/Espo/Core/SelectManagers/Base.php
@@ -825,6 +825,14 @@ class Base
                 $where['value'] = [$from, $to];
 
                 break;
+            case 'beforeXDays':
+                $where['type'] = 'before';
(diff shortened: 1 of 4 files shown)


yurikuzn:
It will be useful filters. Thank you.

alasdaircr:
Never made it in to 4.3 ?

yurikuzn:
Maybe in 4.4.

yurikuzn:
"Older than X days" is understandable name. But no clue how how to name next days filter.

alasdaircr:
I know the names are tricky to figure out so they are understandable - but they are queries which you can't do right now. Maybe time to look at all the options and try and rationalize?

PR #311, opened 2016-11-19
Duplicate record checker in create mode

alasdaircr:
This isn't completely bug free I think but it's a useful feature. Maybe you would like to take something like this on?

The benefit is the user's get notification of duplicates whilst entering the details, not afterwards.

When entering a new Lead, Contact or Company's name,  an ajax request is made and returns possible matches. If a match is clicked, user is taken to that record instead.

In the case of creating a new lead with accountName == an existing Account, create a  Contact instead.

In the image you see a company name matching existing records. In this case a Contact would be preferable rather than a lead.

In order to use the existing suggestion code, the dual first + last name input fields are replaced with one field while in create mode. First and Last are split from the concatenated name when the record is saved.

Also -jQuery.autocomplete modified so that it supports multiple simultaneous request. This is so you can query Leads + Contacts and wait on both results before presenting to the user.

Screen shot of how it looks for us:
2016-11-18_563x547_escrotum

### DIFF
--- a/application/Espo/Modules/Crm/Resources/metadata/entityDefs/Account.json
+++ b/application/Espo/Modules/Crm/Resources/metadata/entityDefs/Account.json
@@ -2,6 +2,7 @@
     "fields": {
         "name": {
             "type": "varchar",
+            "view": "Fields.AccountName",
             "required": true,
             "trim": true
         },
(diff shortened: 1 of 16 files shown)


alasdaircr:
Fixed the conflicts

bandtank:
@alasdaircr Can this be extended to check for duplicates of any created object relatively easily? I would like to use this exact functionality in many places, e.g. addresses, client (I renamed accounts to clients) names, contact names, phone numbers, email addresses, etc.

alasdaircr:
@bandtank potentially yes, but only once it's in core espo.

sab0tajue:
Yes, that would be a great solution.

alasdaircr:
No longer use this application

PR #331, opened 2016-12-07
stream note: use datetime instead of datetime-short

ayman-alkom:
Ex: Yesterday 14:35 instead of just : Oct 10


yurikuzn:
You can see full date on mouse hover.

ayman-alkom:
:) OK

PR #351, opened 2017-01-04
Compose from email address not sorted by primary

alasdaircr:
The compose email dialog uses the user's email address with the lowest id (first in table), rather than the primary one.

This fixes it but may not be they way you'd do it.


yurikuzn:
Thanks. I thought it was implemented.

PR #352, opened 2017-01-04
Hide converted leads from Global search

alasdaircr:
I don't think users really need to have both the Contact & the converted Lead record in the Global Search results.


yurikuzn:
I need it when I use crm some times. 

p.s. tab should be 4 spaces

PR #353, opened 2017-01-04
Show account name in global search results

alasdaircr:
I suggested this years ago, but it wasn't chosen. Trying again :-)

Seems useful that with a large contact database, the global search displays not only the matching contact name but the account name too.


yurikuzn:
Note: Module names should be formatted like "views/global-search/account-field" instead of "Views.GlobalSearch.AccountField" that is deprecated.

alasdaircr:
@yurikuzn have pushed that change

alasdaircr:
So shall I presume that this isn't needed?

yurikuzn:
Not needed as it is now. Maybe we will improve it in more elegant way.

PR #369, opened 2017-01-16
Role Inactive support for Contact

alasdaircr:
Hi - I mentioned this in the forums a while ago.

http://forum.espocrm.com/forum/developer-help/5919-inactive-role-feature-for-contact-account-relationship

It's a useful way of tracking people who have left an account.

Modifies the account_contact table to include a boolean field `active`, and then modifies default behaviour so that contacts with an inactive role are not retrieved when being queried in relation to an account.

However I think it could be cleaner. Im not an expert in the internals of the ORM and the way I'm handling the `accountActive` field in Base.php is not great. Would welcome any suggestions and potential inclusion in core Espo.

There's also some strange behaviour where using the Reports feature, filtering using Account works, but not Accounts. This will be down to my addition of another join mentioned above.

### DIFF
--- a/application/Espo/Modules/Crm/Repositories/Contact.php
+++ b/application/Espo/Modules/Crm/Repositories/Contact.php
@@ -41,9 +41,13 @@ class Contact extends \Espo\Core\ORM\Repositories\RDB
             $params['customJoin'] = '';
         }
 
+        //FIXME This avoids regular requests for Contacts without linked Account properties to work. Must be another solution though?
+        if (!in_array('account', $params['leftJoins'])) {
+            $params['leftJoins'][] = 'account';
+        }
         $params['customJoin'] .= "
             LEFT JOIN `account_contact` AS accountContact
-            ON accountContact.contact_id = contact.id AND accountContact.account_id = contact.account_id AND accountContact.deleted = 0
+            ON accountContact.contact_id = contact.id AND accountContact.account_id = account.id AND accountContact.active = 1 AND accountContact.deleted = 0
         ";
     }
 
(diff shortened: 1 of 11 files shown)


alasdaircr:
Have updated this with my latest version. Now filtering using both Account (Primary) and Accounts is working correctly (the inactive Contact is not returned).

This requires a patch to Advanced Pack which I have available if you're interested.

yurikuzn:
We're going to utilize this. Thanks.

alasdaircr:
Wonderfull news. I'm sure you'll come up with a cleaner solution! It was not easy to get this to work for me.

One thing I thought of recently was I should remove account_id (Primary Account) link from Contact when all Accounts are Inactive. That would reduce the number of changes needed made.

Also to have the ability to filter by ContactActive - in other words see previous employees of an Account would be really useful.

alasdaircr:
Patch which clears primaryID when all account links are inactive, and fix a typo function call.

```
diff --git a/client/modules/crm/src/views/contact/fields/accounts.js b/client/modules/crm/src/views/contact/fields/accounts.js
index 1c7fd0b..118415e 100644
--- a/client/modules/crm/src/views/contact/fields/accounts.js
+++ b/client/modules/crm/src/views/contact/fields/accounts.js
@@ -71,6 +71,8 @@ Espo.define('crm:views/contact/fields/accounts', 'views/fields/link-multiple-wit
                     if (otherActive.length) {
                         let $newPrimary = this.$el.find('button[data-action="switchPrimary"]:first');
                         this.$el.find('button[data-action="switchPrimary"]:not(.active):first').click();
+                    } else {
+                        this.setPrimaryId(null);
                     }
                 }

@@ -198,7 +200,7 @@ Espo.define('crm:views/contact/fields/accounts', 'views/fields/link-multiple-wit
                 if (!activeCols.length) {
                     this.$el.find('button[data-action="switchPrimary"]').removeClass('active').children().addClass('text-muted');
                     $primary.addClass('active').children().removeClass('text-muted');
-                    this.setPrimaryId(id, false);
+                    this.setPrimaryId(id);
                 }
             }
         },
```

PR #397, opened 2017-02-07
remove deprecated config, and fix query search

ayman-alkom:


yurikuzn:
This feature was implemented for those who would like text filters to search using contains operator. They can add the parameter to config manually.

ayman-alkom:
yes I now understand

PR #483, opened 2017-04-06
Excel exports

alasdaircr:
Hello,

See http://forum.espocrm.com/forum/developer-help/25889-excel-exports.

There's patches to Advanced Pack which I can share with you if you're interested in this. Most of the hard work is done - you could chose to offer a choice to users of CSV / and or Excel and maybe reinstate the Export fields modal window which I disabled in favour of saving a sane set of export fields for all users. (Right now CSV export is not that useful apart from machine reading).

Advanced Pack Report sending could gen an excel file and send it rather than the HTML email table too.

exportlayout

tasks

### DIFF
--- a/application/Espo/Core/Utils/Layout.php
+++ b/application/Espo/Core/Utils/Layout.php
@@ -107,6 +107,24 @@ class Layout
             $defaultPath = $this->params['defaultsPath'];
             $fileFullPath =  Util::concatPath(Util::concatPath($defaultPath, 'layouts'), $name . '.json' );
 
+            if ($name === 'exportFields') {
+
+                $listLayoutPath = Util::concatPath($this->getLayoutPath($scope, true), 'list' . '.json');
+                if (!file_exists($listLayoutPath)) {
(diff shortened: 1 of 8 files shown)


bandtank:
This is really nice. Thanks for putting this together. Hopefully it makes its way into the codebase because I really need this.

yurikuzn:
Thanks. We need to have CSV export available too. I think we can apply this in 4.7 version.

alasdaircr:
That's great to hear, Yuri. One less patch to maintain. There's probably some strangeness, but you know the code better than I do.

couple of points maybe worth thinking about

- We like the ability to set the 'Export Layout' in Admin, and don't like the Export modal window. I disable it for that reason. If you want a custom export, use Report feature.
- The exporting is a little bit slower than CSV. In order to have proper field exports I perform a lot of processing in ``Record::loadAdditionalFieldsForExport()`` I'm sure this could be reduced. In particular, rather than loading all the related fields, it would be good to load only the related fields based on what is actually being exported (the selected columns).

yurikuzn:
https://github.com/espocrm/espocrm/issues/515

PR #509, opened 2017-04-19
Fix view name resolving for custom field types in merge action

yunga91:
In _FieldManager_, in _def_ object field types are stored with first char in lowercase. So for defined custom field types method _getViewName()_ returned view name without 'custom:' prefix (default behavior for not existing field type in _defs_ object) .
Also fixed argument typo on line 170.


yurikuzn:
Thanks

PR #516, opened 2017-04-24
Fix typo to have correct translation

sebdoido:
In all the current translation files,
```
application/Espo/Modules/Crm/Resources/i18n/<language>/MassEmail.json
```

the key item is called `Send Test`

This PR fixes the view file `client/modules/crm/src/views/mass-email/detail.js` to have a correct translation of this item


yurikuzn:
Thank you.

PR #524, opened 2017-04-27
update diff to use util instead of sys

bandtank:
Node sys changed to util a long time ago. Using sys shows a warning while util doesn't and util is a drop-in replacement for sys.


yurikuzn:
Thanks

PR #525, opened 2017-04-27
en_US Updates

bandtank:
None of this is a big deal, but I made a few edits to make the labels read more naturally in English.


yurikuzn:
Thanks

PR #538, opened 2017-05-11
Fix and improve translations ES

PanuWeb:


yurikuzn:
Thank you.

PanuWeb:
Ready! I regret that my email is not available. But where did he get it from? In my github account the data is correct

yurikuzn:
I found the email in our CRM. Maybe it was not yours.

PR #546, opened 2017-05-18
Fix view name resolving for custom field types in filter action

yunga91:
Changed custom field type view resolving using FieldManager in filter view.


yurikuzn:
Thanks

PR #552, opened 2017-05-29
Create Lead from input email on "Group Email Accounts"

ZinitSolutionsGmbH:
In "Group Email Accounts" adding checkbox.  
If its enable all email put into Lead. 

The field "email" is filled in "from" fo email.
The field "description" is filled in "body" of email
The field "firstName" is filled in if we have "fromString".


yurikuzn:
Hi,

Thanks for your PR. We won't accept it. We are going to implement such a feature in future.

PR #553, opened 2017-05-29
Refactor link-multiple-with-role to support multiple columns + allow filtering

alasdaircr:
Refactor link-multiple-with role so that the 'role' field isn't hard coded, and multiple columns are supported.

Only enum fields are supported in search mode. Varchar + enum are supported in edit + detail mode.

Would require some better styling for > 2 columns.

screen shot 2017-05-26 at 17 21 06

screen shot 2017-05-26 at 17 21 39


alasdaircr:
This has a bearing on the Inactive Role PR #369. I'm not sure the best way to handle this. Have you made any progress with that patch?

yurikuzn:
Not yet. Maybe I will tackle it for the next version.

alasdaircr:
That would be nice from my POV > one less patch to maintain.

This PR isn't dependent on #369 by the way, it just breaks #369.

alasdaircr:
Currently there's a lot of duplication between `crm:views/contact/fields/account` and `views/fields/link-multiple-with-primary`.

alasdaircr:
With the suggested changes, `views/fields/link-multiple` would handle all of this

yurikuzn:
I didn't have a chance to look into, I'd wouldn't touch `link-multiple` field view.

alasdaircr:
I can have link-mutliple-with-role be an empty view that inherits link-multiple-with-roles/columns in order to maintain backwards compatibility.

But I think my other comments are still valid. The duplication between `link-multiple-with-primary & `contact/fields/accounts` ?

Have a look at this patch at your leisure. I think this is an improvement as it introduces flexibility to the field, and removes some hard coded concepts. I'm going to have to make a hack in order to keep 'inactive' roles working for us, but won't do anything major until you have feedback.

yurikuzn:
You have a hack using 'columns' parameter. that I would make another way. If I implement it I will do it another way.

alasdaircr:
You mean defs.columns?  didn't feel like a hack when I was doing it. The columns were already there just hardcoded as 'role'.

yurikuzn:
We stick to having only 3 parameters: type, attribute, value.

alasdaircr:
Ok I see what you mean. Yes that's certainly not the greatest part of the patch.

Remember I come from a position of not writing the code, and not understanding why users aren't able to filter by things which are stored in the database. For instance - the enum roles.

yurikuzn:
I was going to make it possible to have fields like `opportunityRole` available as a separate filter of enum view.

alasdaircr:
I'd also considered that, don't remember why I did it this way instead. Looking forward to seeing it then, and putting this patch in the bin!

yurikuzn:
When I proceed to this feature I will investigate this more deeply.

PR #565, opened 2017-06-14
Fix #563: Replace Container::loadLog() with loader class

### LINKED ISSUE #563: Replace all loadSmth-methods with loaders to get an ability to override them by custom loaders
## Problem
I want to enable error monitoring for my EspoCRM installation by Rollbar service.
The integration is pretty simple: I just need to push additional handler into the internal `Monolog` class.
But all of the classes (`Application`, `Container`) are harcoded - and I can't inject my custom workaround as a new dependency.

## Reason
There is no way to extend the way it works for loadSmth-methods (e.g., `loadLog`). In other words, I can't integrate it with Rollbar without hacking.

## Solution
Let's turn all loadSmth-methods (including `loadLog`) into `\Espo\Core\Loaders` classes. It will give an ability to override them by creating the corresponding class in `\Espo\Custom\Core\Loaders` namespace (e.g., `\Espo\Custom\Core\Loaders\Log`) with required tweaks.

PatchRanger:
As I've mentioned, it would fix #563.
The only problem I see with such approach - is that there could be child Conatiners which extend core Container and rely on loadLog() method (e.g., calling `parent::loadLog();`). I did a quick search: none of the existing containers use loadLog such way. So it looks safe to make such change.

### DIFF
--- a/application/Espo/Core/Container.php
+++ b/application/Espo/Core/Container.php
@@ -93,31 +93,6 @@ class Container
         return $className;
     }
 
-    protected function loadLog()
-    {
-        $config = $this->get('config');
-
-        $path = $config->get('logger.path', 'data/logs/espo.log');
-        $rotation = $config->get('logger.rotation', true);
-
-        $log = new \Espo\Core\Utils\Log('Espo');
-        $levelCode = $log->getLevelCode($config->get('logger.level', 'WARNING'));
-
-        if ($rotation) {
-            $maxFileNumber = $config->get('logger.maxFileNumber', 30);
-            $handler = new \Espo\Core\Utils\Log\Monolog\Handler\RotatingFileHandler($path, $maxFileNumber, $levelCode);
-        } else {
-            $handler = new \Espo\Core\Utils\Log\Monolog\Handler\StreamHandler($path, $levelCode);
-        }
-        $log->pushHandler($handler);
-
-        $errorHandler = new \Monolog\ErrorHandler($log);
-        $errorHandler->registerExceptionHandler(null, false);
-        $errorHandler->registerErrorHandler(array(), false);
-
-        return $log;
-    }
-
     protected function loadContainer()
     {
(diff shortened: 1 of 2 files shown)


yurikuzn:
Could you accept our CLA before we could accept you PRs?
https://github.com/espocrm/cla

PatchRanger:
@yurikuzn Here you are: https://github.com/espocrm/cla/pull/12 .

yurikuzn:
This change caused a serious problem. One of them: https://github.com/espocrm/espocrm/issues/630

We are going to fall back this change.

PatchRanger:
@yurikuzn Thank you for notification. Please make sure the PR is really related - as it neither adds new functionality nor breaks the way it is working, it just switches from one way of autoloading to another. In other words it shouldn't (even in theory) cause any trouble. The only case for #630 I could imagine: upgrade was not complete (missing class file) and/or not valid cache or registry.

yurikuzn:
It's the one example. We run into the problem a few times since this change.

PatchRanger:
What is the exact reason of the problem? Is there any other problems with `Espo\Core\Loaders`-loading? According to https://github.com/PatchRanger/espocrm/blob/master/application/Espo/Core/Container.php#L70 `loadLog` and `Espo\Core\Loaders\Log` are totally equal - except the case of invalid metadata. Please investigate the root of the issue, do not fight with consequences - please do not rollback this PR until the real reason of the problems is discovered. It seems it is related to metadata - I guess, just `Registry rebuild` is enough.

yurikuzn:
Metadata is loaded before log. In case of any notice/warning/log the script will end.

PatchRanger:
> Metadata is loaded before log.

So the aforementioned line of code should work as intended.

Are you sure reverting this change will fix other issue? The root reason still looks strange for me, I don't understand what's causing the issue.

Issue #446, opened 2017-03-06
Continuous deployment (build process) setup


tannakartikey:
Once we clone the repo we can work in the development environment and generate the build.

Once the build is deployed on the production, how do we deploy next changes to the server without breaking the code or losing the data?

I would like to set up the staging and the production and frequently would like to import the data from the production to the staging.

I could not find anything regarding this in the documentation or on the forum.

yurikuzn:
It's a good question. Since we usually have some scripts in upgrade packages I'd suggest using our upgrades. But there can be problems if you have modifications in core files

tannakartikey:
Thanks for the answer @yurikuzn
Do we have them documented in any form or I'm supposed to go through the code to understand them?

I would like to work towards it since it is an essential part of the SDLC

yurikuzn:
Seems I missed the point. I meant to upgrade your inctance in regular way using Administration > Upgrades. That's only only way we support.

@alasdaircr, maybe you can give some advices? I believe you have experience with this. Thanks.

bandtank:
@tannakartikey are you trying to deploy to an HA environment? If yes or if you're trying to deploy to any environment that depends on a stateless server, it won't work. EspoCRM was not designed to scale horizontally to an HA or stateless environment. I'm working on a fork of the repository that will fix this issue and basically turn it into a more commercial product, but it's going to be a while. Just FYI.

What you're trying to do is a really typical build and deploy process, but you have to do it in place on the same server each time with downtime. As I mentioned above, all of the uploads, preferences, and whatever files you store in the application will end up on a disk instead of at an object store like S3. 4.5.0 introduced a fileManager class that is supposed to help move objects off of the local server, but there are many more issues that prevent a typical app deployment. This also makes it basically impossible to make a reliable backup of the local files, which is the primary and major issue with how files are currently handled.

The short answer is plan for downtime and try to deal with the consequences of the server maintaining your files and application's state. There's no way to run this application through a build pipeline because of all the dependencies that aren't saved outside of the actual application's installation directory. If you find a way to do this, please let me know. As I said, I'm working on it right now because the rest of the application is good, but it can't be used in any kind of scaled or HA environment.

Note: you can deploy the data/config.php file with a typical deployment process, so if you can figure out how to handle all of the other types of files that need to exist locally, then you can do a real deployment.

tannakartikey:
@bandtank no, I am not trying to scale it up. Planned downtime is also fine with me.

[...]

PR #580, opened 2017-07-03
Fix #567: Add custom config override

### LINKED ISSUE #567: Add environment-specific configuration overriding defaults in global configuration and ignored by git
## Problem
There is no way to have global configuration with reasonable defaults (which is versioned by git) - together with environment-specific tweaks (which are ignored by git).

## Reason
For now we do not have a mechanism of including (if exists) environment configuration, which overrides global values.

## Solution
We should adapt something similar to [what Yii currently has](https://github.com/yiisoft/yii2-app-advanced/blob/master/docs/guide/structure-environments.md).

PatchRanger:
### Problem
See #567.

### Solution
Just add a way to override global configuration by adding a `data/customConfig.php` file.
It solves a problem of having different environment-specific settings: each environment just could have its own `data/customConfig.php` file, basing on common `data/config.php` as reasonable defaults.
Also it could serve as little step toward #446.


yurikuzn:
Hi,

What is the benefit of having customConfig.php with config.php being in gitignore?

yurikuzn:
Configuration changed through UI will be written in config.php that can be overriden by customConfig.php. So Admim user won't be able to change some parameters.

yurikuzn:
I think properly would be to add defaultConfig.php (verisioned by git) loading before existing config.php.

PatchRanger:
> What is the benefit of having customConfig.php with config.php being in gitignore?
`customConfig.php` should be moved into the correct place during the process of deployment.

> Configuration changed through UI will be written in config.php that can be overriden by customConfig.php. So Admin user won't be able to change some parameters.

That's true. It's a downside of this MR, I agree. Though I don't think there is a huge desire to change DB settings through UI after successful installation. While introducing such change would allow to have different settings for different environments - which is very useful for Continuous Integration.

> I think properly would be to add defaultConfig.php (verisioned by git) loading before existing config.php.

That's an interesting approach. Should we include it by `loadConfig` - or just `require` (if it exists, of course) directly in `config.php`?

yurikuzn:
Including defaultConfig.php file in config.php is a simplest way. I like it.

PatchRanger:
@yurikuzn I found that it's not that easy to just add `include` into the `config.php`. As `config.php` gets rebuilded - the correct place to introduce the change is `Config::save()`. But it's not enough to just add `include 'defaultConfig.php';` above the settings - we need to get the content of that file and merge it with current.
So I decided to just modify the previous implementation: now it considers custom configuration as default settings - not overrides. I guess, it should fit.

PatchRanger:
@yurikuzn Could you please review the proposed changes?

PR #583, opened 2017-07-06
keep seconds part of the DateTime value

barwi:
Prevent to set seconds to zero when hours & mins values are not changed.


yurikuzn:
Hi,

Can you describe how to reproduce the bug.

The fix is not fine to accept. `data` function shouldn't change a state. It's only for displaying data.

barwi:
Time picker in edit mode resets seconds to zero every time, even when no any changes has done on the form. It is not visible on user side because data is formatted by 'hh:mm' template (seconds is hidden). But in general, changing value of the field in this way should be avoided.

PR #584, opened 2017-07-06
use list and detail views as side panels from clientDefs

barwi:


yurikuzn:
https://github.com/espocrm/espocrm/commit/2016a1bb0d6624a0b0146e8b164897318e76c0e9
https://github.com/espocrm/espocrm/commit/6841ba9c85e47a42825d21a6a5c9e2e16a1f5184

PR #586, opened 2017-07-07
Fix #585: Verbose JSON::decode()

### LINKED ISSUE #585: Json::decode():unexpected character
### Problem
I keep receiving such messages into the EspoCRM logs:
```
[2017-07-07 01:07:13] Espo.ERROR: Json::decode():unexpected character [] []
```
There's for sure some issue with JSON-encoding in my installation of EspoCRM - but the error message doesn't give a chance to understand where.

### Proposal
Let's make this error more verbose.

PatchRanger:
### Solution
Taking `JSON::encode()` as a basement and an example, added the value, that breaks JSON-decoding, to the log message.

### DIFF
--- a/application/Espo/Core/Utils/Json.php
+++ b/application/Espo/Core/Utils/Json.php
@@ -68,14 +68,14 @@ class Json
             return false;
         }
 
-        $json = json_decode($json, $assoc);
+        $value = json_decode($json, $assoc);
 
         $error = self::getLastError();
         if ($error) {
-            $GLOBALS['log']->error('Json::decode():' . $error);
+            $GLOBALS['log']->error('Json::decode():' . $error . ' - ' . print_r($json, true));
         }
 
-        return $json;
+        return $value;
     }
 
     /**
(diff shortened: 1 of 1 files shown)


yurikuzn:
Hi,

This might make log files very large.

PatchRanger:
@yurikuzn I agree, but it will make it possible to fix it. Otherwise we have no chance)

PatchRanger:
@yurikuzn `JSON::encode()` does the same thing - it logs `$json` variable in case of fail. I guess, it's required to (possibly) crowd logs in order to notify about the cause of the trouble.

yurikuzn:
I think we could cut the print to the certain number of characters.

PatchRanger:
@yurikuzn I've enabled "Allow edits from maintainers" option - could you please add cutting to the certain number of characters the way you consider appropriate & merge the PR?

PatchRanger:
@yurikuzn I've added cutting to 255 characters, please review.

PR #591, opened 2017-07-12
Fix #590: Avoid IMAP error by unsetting recent flag

### LINKED ISSUE #590: "cannot set flags, have you tried to set the recent flag or special chars?"
### Problem
```
[2017-07-10 11:56:16] Espo.ERROR: EmailAccount 58b7090e3644ee829 (Get Message w/ parser PhpMimeMailParser): [0] cannot set flags, have you tried to set the recent flag or special chars? [] []
```

PatchRanger:
### Reason
IMAP can't handle setting `FLAG_RECENT`.

### Solution
Avoid passing `FLAG_RECENT` to `setFlags()`.

### DIFF
--- a/application/Espo/Services/EmailAccount.php
+++ b/application/Espo/Services/EmailAccount.php
@@ -310,6 +310,8 @@ class EmailAccount extends Record
 
                     if ($emailAccount->get('keepFetchedEmailsUnread')) {
                         if (is_array($flags) && empty($flags[Storage::FLAG_SEEN])) {
+                            // Recent flag may not be set.
+                            unset($flags[Storage::FLAG_RECENT]);
                             $storage->setFlags($id, $flags);
                         }
                     }


yurikuzn:
Did the error stop to occur after your fix? Was the error caused 100% by recent flag?

PatchRanger:
> Did the error stop to occur after your fix?

@yurikuzn Actually I didn't check it as it is quite hard to reproduce the issue manually - without unwanted consequences.

> Was the error caused 100% by the recent flag?

The error message is totally the same as in `zend-mail`: https://github.com/zendframework/zend-mail/blob/master/src/Storage/Imap.php#L530 . It is the only `setFlags()` call through out the code - so it looks like it is the only part affected.

yurikuzn:
It's just a message in zend. But the reason of the error can be any else.

PR #593, opened 2017-07-14
Fix #592: Replace undefined '$email' with '$this'

### LINKED ISSUE #592: Fatal error when using getToList(), getCcList(), getBccList() or getReplyToList()
### Problem
I've got fatal error, when I try to use one of these methods in `Espo\Entities\Email`:
- getToList()
- getCcList()
- getBccList()
- getReplyToList()

### Reason
The reason is obvious: it uses undefined variable `$email` - I guess, `$this` is implied. I am going to provide PR in a moment.

PatchRanger:
### Solution
It was simple - just replace the variable.

### DIFF
--- a/application/Espo/Entities/Email.php
+++ b/application/Espo/Entities/Email.php
@@ -124,7 +124,7 @@ class Email extends \Espo\Core\ORM\Entity
 
     public function getToList()
     {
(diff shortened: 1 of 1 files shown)


yurikuzn:
Thanks

PR #604, opened 2017-08-08
delete unneeded function

ayman-alkom:
even not valid


yurikuzn:
This function is useful and duplicates the same function from the backend.

https://github.com/espocrm/espocrm/commit/f1bf0b3dee1676eda025f698cdde8a9c4000ad9e

PR #623, opened 2017-08-17
Fix clear cache typo update fr lang

sebdoido:
2 commits

Update
client/res/templates/clear-cache.tpl 
client/src/views/clear-cache.js      
to fix typo

+ update fr_FR lang


yurikuzn:
Thanks

yurikuzn:
Target.json is not needed. It has been removed from repository.

sebdoido:
Ok, i think we should update the POEditor records because it has been generated after i downloaded the latest .po file

yurikuzn:
I have just done it.

sebdoido:
Thanks :)

PR #624, opened 2017-08-18
fix today, on, between conditons for DateTime

barwi:
Between operator includes both sides of the range. Conversion to UTC must be applied for each local date.


yurikuzn:
Hi,

Thank you. I've done the same w/o creating additional DateTime instance.

https://github.com/espocrm/espocrm/commit/3120afb55e8e1efe1e0d5647d38197c9a78bcb95

PR #626, opened 2017-08-19
Dashlet: handle header option for calendar

sebdoido:
Currently, when using calendar dashlet, it's not possible to navigate through weeks.. months... 
in `dashlets/calendar.js`, header is set to false

End-user should be able to configure that option and decide if they want to display header or not.

I update `i18n/{en_US,fr_FR}/DashletOptions.json` file but i think POEditor entries should be updated?

Thanks for your support and review!

### DIFF
--- a/application/Espo/Modules/Crm/Resources/metadata/dashlets/Calendar.json
+++ b/application/Espo/Modules/Crm/Resources/metadata/dashlets/Calendar.json
@@ -8,6 +8,11 @@
                 "type": "varchar",
                 "required": true
             },
+            "header": {
+                "type": "bool",
+                "default": false,
+                "required": true
+            },
             "autorefreshInterval": {
                 "type": "enumFloat",
                 "options": [0, 0.5, 1, 2, 5, 10]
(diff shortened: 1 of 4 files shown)


sebdoido:
Here is how it could be displayed "with header", "without header", "options form"

without header
options
with header

yurikuzn:
Hi Sebastien,

I'm not sure that we need an ability to switch view modes in calendar dashlet. But an ability to go switch through the next/previous date ranges is what is good to have. I will think about it. Thank you.

sebdoido:
Hi Yuri

do you think the better option would be to add a "navigation" variable to feature flip navigation buttons... and not display the complete header?

Let me know how you would like to implement that.. so maybe i can try to do it.
I would like to avoid to have a custom patch on my project waiting for the target solution :)

yurikuzn:
I consider about displaying buttons on dashlet header.

yurikuzn:
Done this in https://github.com/espocrm/espocrm/issues/629

PR #631, opened 2017-08-23
CronManager: set warning loglevel if too frequent execution.

sebdoido:
set warning loglevel if too frequent execution.


sebdoido:
up :)

yurikuzn:
I'm not sure whether we need it. In future planned to have an ability to run cron more frequent that once per minute.

Issue #633, opened 2017-08-29
SQL error: Table 'workflow' was not locked with LOCK TABLES


ayman-alkom:
This error occurs when I want to create an entity that has a number type field.
The problem is [here](https://github.com/espocrm/espocrm/blob/master/application/Espo/Hooks/Common/NextNumber.php#L73)

I commented the query sql lock  lines and it worked well, how can we solve this?

ayman-alkom:
I solved it and opening a PR

PR #634, opened 2017-08-29
fix lock tables conflict with workflow table

ayman-alkom:


ayman-alkom:
ref: https://github.com/espocrm/espocrm/issues/633
The problem is that we call the save before release the tables, and the workflow is called normally somewhere in the save operation.

yurikuzn:
This commit should have fixed the problem: https://github.com/espocrm/espocrm/commit/89e13b0bac673cd30b188aaf2d3924aab97d63f4

The purpose of locking table is preventing occuring same numbers. You fix breaks this logic.

ayman-alkom:
yes it have fixed the problem, thanks.
but does it make sense to release the lock before saving the entity like I made in this PR?

PR #637, opened 2017-09-02
Duplicate: re-create related links instead of just update the relation

ayman-alkom:
Hi @yurikuzn 
In my project this change was very important, and I hope you accept it.
If you want we may add a boolean member variable to the record service that enable or disable the re-creation

### DIFF
--- a/application/Espo/Services/Record.php
+++ b/application/Espo/Services/Record.php
@@ -1757,7 +1757,14 @@ class Record extends \Espo\Core\Services\Base
         foreach ($this->duplicatingLinkList as $link) {
             $linkedList = $repository->findRelated($duplicatingEntity, $link);
             foreach ($linkedList as $linked) {
-                $repository->relate($entity, $link, $linked);
+                $arr = $linked->toArray();
+                $foreignIdName = lcfirst($this->entityType) . 'Id';
+                unset($arr[$foreignIdName]);
+                unset($arr['id']);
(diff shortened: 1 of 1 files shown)


ayman-alkom:
Hi yuri
any feedback for this

yurikuzn:
Hi Ayman,

I don't think we need to have this feature. It's possible to come up with a lot of interesting features that would be needed for some, but it all inflates codebase.

rodrigoscoelho:
Just asking, please don't badly understand me. Is espocrm based on an open
community? Or it is a closed app with some code open to allow users to fork
and create their own versions?

I'm not a programmer, and I'm not experienced on this, so, forgive me if
I'm asking something obvious.

Anyway, it's a good piece of code....



Em 6 de set de 2017 14:59, "Yuri Kuznetsov" <notifications@github.com>
escreveu:

yurikuzn:
It's open source. GPLv3. You can fork an entire application.

Some additional extensions like advanced pack are not under GPLv3. They have their own licenses and you not able to redistribute.

ayman-alkom:
sometimes I feel the same as @rodrigoscoelho do, I feel like this app is not community driven project, and in the same time I understand your opinion that some features may reduce the code safety.
I feel that some feature like this is strongly wanted, and easy to be done.
any way you can can close this PR, I can keep it in my own project only.

PR #661, opened 2017-10-14
list layout improvements

ayman-alkom:
1 - I noticed that the layout is converted every time we call build row 
the reason is that forEach is slower than getting the layout 
I know it is not important change but liked to share it

2 - apply ACL on the returned layout object whereas just passed the readable ones
so the columns will disappear

### DIFF
--- a/client/src/views/record/list.js
+++ b/client/src/views/record/list.js
@@ -999,10 +999,21 @@ Espo.define('views/record/list', 'view', function (Dep) {
                 return;
             }
             this._loadListLayout(function (listLayout) {
-                this.listLayout = listLayout;
-                this._internalLayout = this._convertLayout(listLayout);
+
+                var scopeForbiddenFieldList = this.getAcl().getScopeForbiddenFieldList(this.entityType, 'read');
+                var _listLayout = [];
+
+                for (var i in listLayout) {
+                    var col = listLayout[i];
+                    if (scopeForbiddenFieldList.indexOf(col.name) !== -1) {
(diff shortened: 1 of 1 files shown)


ayman-alkom:
hi @yurikuzn 
we can cancel the first commit if you didn't like it, and discuss just the second one
or we can discuss the both
I tested them well, and no problem
and found them useful so liked to share them with you
normally I made more changes but not share it because I know that you will not accept them

yurikuzn:
Hi,

List in History panel with emails, meetings and calls having different layout for emails.

yurikuzn:
Initially layout was being loaded before: https://github.com/espocrm/espocrm/blame/f062c08b31f6811a651fa5f416cd5ee99e19deb2/frontend/client/src/views/record/list.js#L637 but I needed to change it so that it's loaded per each row. No problems with performance. Data is cached.

ayman-alkom:
getInternalLayout function will be called 20 times in the first page item, and then will be returned from the cache.
this is not very important, but in the same time a little well.

yurikuzn:
Regarding field level ACL I will investigate it when get a chance.

ayman-alkom:
I hope this change is ok now, I removed that commit and kept the ACL applying one

yurikuzn:
I will investigate it and will apply this logic for the next release if everything will be ok. Thanks.

https://github.com/espocrm/espocrm/issues/669

ayman-alkom:
👍

PR #666, opened 2017-10-19
Htmlizer: handle hasMany, hasChildren, manyMany relations (not only b…

sebdoido:
Currently, only belongsTo and belongsToParent relation are handled to get entityData.

Through this commit, it's now possible to load other relations data (like hasMany) to use in template {{#each}} iteration

Feel free to comment or amend the way you would like to implement that :)
Thanks for your feedback

### DIFF
--- a/application/Espo/Core/Htmlizer/Htmlizer.php
+++ b/application/Espo/Core/Htmlizer/Htmlizer.php
@@ -81,7 +81,7 @@ class Htmlizer
         return $value;
     }
 
-    protected function getDataFromEntity(Entity $entity, $skipLinks = false)
+    protected function getDataFromEntity(Entity $entity, $skipLinks = false, $relationFields = array())
     {
         $data = $entity->toArray();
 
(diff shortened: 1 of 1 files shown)


yurikuzn:
What if Account have 10,000 calls.

sebdoido:
You're completely right.
I've added a new argument to render() method (with default value to array()) to be able to specify which relations we would like to load..

Then, checking if the $relation is in_array of expected fields...
Let me know what you think about it or how you would like to implement that ?
Thanks Yuri!

yurikuzn:
There is no determined order in which related records will be listed. Since relation list must be passed as a method argument it's not usable in an existing print pdf feature.

I would rather use $additionalData  argument with the prepared list.

sebdoido:
I rollback what i proposed and tried to use `$additionalData` as you suggested. 
but .. i have the feeling that I should reuse most of the code present in 

`protected function getDataFromEntity(Entity $entity, $skipLinks = false)` 
to prepare parse the data from my many-to-many relations... 

(iterate over the keys... test `if ($type == Entity::DATETIME) { ` to display a proper datetime.. (`convertSystemDateTime`) and so one... 
I would like to avoid to have to re-implement that on my side just to call the `render` function with a prepared `$additionalData` variable well-formed... :-(

any suggestion ?
Finally, i think that adding an extra args like i propose before is maybe not a bad way of doing it ?

Let me know what you think about Yuri! Thanks for your kind help!

yurikuzn:
I'd re-implement. Don't see a problem there.

sebdoido:
Ok, but it will be duplicated code :-( 
maybe i will do a custom patch on your file that i will maintain with my stuff :)

PR #667, opened 2017-10-19
Entrypoints: adding getTemplateFileManager method

sebdoido:


yurikuzn:
Hi,

No need to have this method in Base class.

PR #689, opened 2017-11-13
Route: possibility to replace routes in modules

ZinitSolutionsGmbH:
This modification will allow to redefine the routes in the modules. That is, each subsequent loaded module can change the controller that processes the request. It gives more flexibility to the system, because very often one module extends the previous one, etc.

### DIFF
--- a/application/Espo/Core/Utils/Route.php
+++ b/application/Espo/Core/Utils/Route.php
@@ -112,17 +112,27 @@ class Route
         }
     }
 
+    /**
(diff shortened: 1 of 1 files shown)


tmachyshyn:
Please write tests with core files modifications, https://github.com/espocrm/espocrm/blob/master/tests/unit/Espo/Core/Utils/RouteTest.php.

ZinitSolutionsGmbH:
ok

PR #691, opened 2017-11-14
Removed loaders from Container

ZinitSolutionsGmbH:
Refactored of container class. Removed loaders from container (except metadata). It gives more flexibility for the system. With this feature we can replace or extend loaders in modules.

### DIFF
--- a/application/Espo/Core/Container.php
+++ b/application/Espo/Core/Container.php
@@ -1,5 +1,5 @@
 <?php
-/************************************************************************
+/* * **********************************************************************
  * This file is part of EspoCRM.
  *
  * EspoCRM - Open Source CRM application.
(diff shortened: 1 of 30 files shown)


yurikuzn:
This will cause some serious issues.

ZinitSolutionsGmbH:
What issues do you mean? Everything works fine in my project.

yurikuzn:
Notice error, exception occurred before log is loaded will cause fatal error. Log loading shouldn't depend on metadata.

We accepted such a PR before and had a problem: some user could not upgrade their EspoCRM because of a single NOTICE occurring due some minor issue.

zsadmin:
I agree with you. But what about the rest of the loaders? If I return Log loader to container, you accept pull request?

I can explain my point. I tried to modify your entity filters and I found the class that filters all entities Espo\Core\SelectManagers\Base. For extending this class I should change \Espo\Core\SelectManagerFactory but this factory is loaded in container. 

In Container class I saw a lot of loaders that can be moved to files, and this will solve difficulties in future.

yurikuzn:
I can't accept this by now. This is something that needs time to investigate. I don't have time for now.

yurikuzn:
You can create custom select manager for all your entity types.

zsroma:
I know. But i should create a files for all entities (copy-paste). That's why i find another way to solve a problem.

Thank you for anwer. 

I will prepare a new request without Log loader.

PR #705, opened 2017-11-21
Date Time: check if timezone has a value

ayman-alkom:
I have just downlaod the last version 4.8.4
and tried to create new document 
then I had this error:

> Uncaught TypeError: Cannot read property 'format' of undefined    at DateTime.getToday

the error fixed bey checking the value of time zone before using it 
note: I am not sure if it is well to use this.getTimeZone() instead of this checking

### DIFF
--- a/client/src/date-time.js
+++ b/client/src/date-time.js
@@ -185,7 +185,11 @@ Espo.define('date-time', [], function () {
         },
 
         getToday: function () {
-            return moment().tz(this.timeZone).format(this.internalDateFormat);
+            var m = moment();
+            if (this.timeZone) {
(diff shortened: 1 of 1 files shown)


yurikuzn:
Hi,

Could  you provide steps to reproduce on not customized espocrm? I can't reproduce.

yurikuzn:
I could reproduce.

yurikuzn:
Fix: https://github.com/espocrm/espocrm/commit/0797b71d9554f5571d7060ff0f84bbd6d2c272b3

ayman-alkom:
ok, using getTimeZone() works too

PR #709, opened 2017-11-23
fix enum translate in list view when linkable

ayman-alkom:
this time it is fix and not a feature request, so I expect you to accept it :)


ayman-alkom:
do you prefer the template to be in one line?

yurikuzn:
Thanks

PR #740, opened 2017-12-22
Opportunity: be able to update stage (closed/lost) from listView

sebdoido:
Hi there

like you did it for "Call" listview (be able to setHeld / setNotHeld), it could be great to be able to update in mass "stage" of opportunity to "Closed Won" / "Closed Lost".

Todo: update PO file to add
setClosedWon
setClosedLost

to be translated

Thanks for your review!


yurikuzn:
Hi,

It can be done with Mass Update.

It doesn't updates probability though, we will fix it.

sebdoido:
Hi Yuri

yes, but it's "many clicks" with Mass Update :-)
I like the way you did it for "Call", and be able to set through Action
dropdown list "Set held / Set not held" without having to go through Mass
Update.

PR #756, opened 2018-01-13
Save data import scheme

Tothewine:
Import data in EspoCRM, don't  save data import scheme.


yurikuzn:
Please could you clarify more detail.

PR #798, opened 2018-02-08
refactor variable name

brianpunzalan:


yurikuzn:
Hi,

We need to remve variable here. And leave just
```
$this->getEntityManager()->getRepository('PhoneNumber')->storeEntityPhoneNumber($entity);
```
Thanks

brianpunzalan:
Yes, it is also the same with line:
https://github.com/espocrm/espocrm/blob/master/application/Espo/Core/ORM/Repositories/RDB.php#L371

yurikuzn:
I've already done it. Thanks.

Issue #833, opened 2018-03-16
My Activities Dashlet Display Count Issue


Sloff:
Hi

The My Activities Dashlet on my Dashboard has the option set to display 10 records, but it's only displaying 5 and then "Show More". The Stream dashlet does not have this issue

yurikuzn:
Thanks for reporting

PR #835, opened 2018-03-16
Bug/833: Activities Dashlet: fix Display Records option

tanyalei:


yurikuzn:
https://github.com/espocrm/espocrm/issues/833

PR #871, opened 2018-04-12
parseEmailToMarkdown function added for stream email posts

Sloff:
This improves the look and feel of emails within the stream


yurikuzn:
Hi,

I can't apply this commit. I'm not sure whether this will function reliably.

Sloff:
Would it change your mind if I add several unit tests for the function?

Sloff:
Or I could simply provide a couple of example emails and how they appear in the stream

Sloff:
There is also a composer package that can accomplish this: https://packagist.org/packages/league/html-to-markdown

yurikuzn:
I don't have time now. I will leave it aside for the future.

PR #969, opened 2018-07-13
relationship view: ability to skip collection select

ayman-alkom:
Hi Yuri
Please look at this change if is applicable or not, because it is sometimes very important to retrieve all the values in custom relationship views


ayman-alkom:
screen shot 2018-07-13 at 2 13 50 pm

ayman-alkom:
thanks, your advice can solve the problem, but may skipping it in the relationship view still make sense because it just affect the relationship state.

PR #1017, opened 2018-08-22
Saml

srdjan-catalyst:


srdjan-catalyst:
Hi Yuri,

On 23/08/18 19:21, Yuri Kuznetsov wrote:
> Hi,
> 
> Thanks for PR. I have a few questions:
> 
> 1. Are modifications on the client side are necessary?

I think so. I believe you have to hit the SP page to get authenticated,
you cannot just pass username/password, so redirection is necessary. But
if you have an idea how to do redirection outside UI I'm happy to go
that way.


> 2. Is there a way to move all configuration parameters to data/config.php?

Yes, however it is a bit harder. I suspect people would be configuring
SimpleSAMLphp SP anyway, so then they would need to copy config.

I also wrote SAML.md with an overview of what I did and how it is used,
but it seems to be lost, it is not in the commit where it is supposed to
be. I'll try to find it, and if not I'll write it again. Then it will be
a bit more clear.

Srdjan
>

hturkoz:
👍

hturkoz:
Dear Yuri, 
I have not skill .. but a SSO (single sign-on) it's a "must have".
I know it's time, is work.. but can you just think about when you have a little time :) 

Regards

tmachyshyn:
@srdjan-catalyst I'm trying to test your pull request. Could you provide more details about your implementation. You wrote that added some SAML.md, but this file is not exists here.

srdjan-catalyst:
Sorry, missed to add it. Done now. Cheers.

hi-ko:
Please keep in mind that SAML authentication may have limited value if we are not able to sync groups (& users) s. #1244
Additionally before spending to much energy into SAML/Espo integration please keep in mind that SAML is only one SSO/Authentication use case from many in the field. Instead of implementing/maintaining/configuring SAML and other authentication use case for every app I would prefer a SSO app in front like LemonLDAP::NG, CAS, Keycloak, Gluu and to implement only the basic additional services which are best practice for SSO integration:
- user & group sync
- login by env variables (HTTP header like 'REMOTE_USER')
- logout

PR #1154, opened 2018-12-19
Allow changing name of the default filter name

barwi:
Default filter name (literally 'All') can be translated for each entity.
It's useful when filter code is changed in SelectManager class.


yurikuzn:
Thanks.

https://github.com/espocrm/espocrm/commit/ae24fc07ba87b0b0160c7e1b58b160c327e6064e

PR #1158, opened 2018-12-20
small fix to activities service

ayman-alkom:
`$selectParams` in the line [L1057](https://github.com/espocrm/espocrm/blob/master/application/Espo/Modules/Crm/Services/Activities.php#L1057) was not defined because of being before `$selectParams` initialized, so I moved it to be after.
`$selectManager->addLeftJoin(['assignedUsers', 'assignedUsers'], $selectParams); `seems not needed as it is added in the line 1095


yurikuzn:
Thanks. 

Fixes are better to push to hotfix/* branch. Master is for the new features.

ayman-alkom:
> Thanks.
> 
> Fixes are better to push to hotfix/* branch. Master is for the new features.

you are right

PR #1234, opened 2019-02-13
 Allows Entry Points of Specific modules to be customized/overridden

laxminarayan-uc:
Currently, Modules can have their own EntryPoints, but they cannot be customized. This patch fixes that.

### DIFF
--- a/application/Espo/Core/EntryPointManager.php
+++ b/application/Espo/Core/EntryPointManager.php
@@ -53,7 +53,7 @@ class EntryPointManager
         'corePath' => 'application/Espo/EntryPoints',
         'modulePath' => 'application/Espo/Modules/{*}/EntryPoints',
         'customPath' => 'custom/Espo/Custom/EntryPoints',
-	'customModulePath' => 'custom/Espo/Custom/Modules/{*}/EntryPoints'
+        'customModulePath' => 'custom/Espo/Custom/Modules/{*}/EntryPoints'
     );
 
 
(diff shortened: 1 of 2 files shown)


laxminarayan-uc:
Forgot to rebase

yurikuzn:
Hi,

It's supposed that entry points of modules will be customized in Custom\EntryPoints\. Doesn't it work?

Customizing in \Custom\Modules\{ModuleName}\ is not supposed by design. I'm not willing to add such an ability w/o being totally sure that it's a really necessary feature.

laxminarayan-uc:
Keeping EntryPoints of different modules in the same folder imposes these problems

1. That makes the directory structure inconsistent with the original directory structure.
1. To keep our Module customizations manageable, we have decided to keep the customizations of each module in seperate repositories. The lack of separation of EntryPoints for modules prevents us from having separate repositories, as they need to go into the same folder .
1. It also becomes an issue when dockerizing for development.  The code for the **customization** of different modules lives in different folders in the host machine and the volumes can be individually mounted to different module folders. Not allowing customization folders of individual modules to have their own EntryPoints folders prevents us from achieving this separation.

yurikuzn:
1. We can't have it only for entry points. If for entry point then for everything else.
2. It's a quite serious change. So I need to think it out before implementing it.

laxminarayan-uc:
Sounds good

yurikuzn:
But I have 200+ items in my todo list, so this is not my high priority. I recommend using just \Custom\EntryPoints\ for now.

laxminarayan-uc:
We can patch my commit before we use it. I was only sharing it with hopes that it will be useful for others.

PR #1247, opened 2019-02-20
list record: enable top bar  if "display total count" is true

ayman-alkom:
I hope this change to be accepted as the normal users are not able to see the totals even it is enabled in the config

Best Regards

### DIFF
--- a/client/src/views/record/list.js
+++ b/client/src/views/record/list.js
@@ -335,6 +335,7 @@ Espo.define('views/record/list', 'view', function (Dep) {
         data: function () {
             var paginationTop = this.pagination === 'both' || this.pagination === true || this.pagination === 'top';
             var paginationBottom = this.pagination === 'both' || this.pagination === true || this.pagination === 'bottom';
+            var displayTotalCount = this.displayTotalCount && this.collection.total > 0;
 
             return {
                 scope: this.scope,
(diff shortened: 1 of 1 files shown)


yurikuzn:
Won't this show top bar in places where it's not supposed to be shown? Like relationship panels, dashlets?

ayman-alkom:
if `displayTotalCount = false` then this change will not affect anything
as I noticed the value is false in the relationship so the effect only will be in the main lists
also I think that this condition is needed as the top bar already include this piece of code (total)

ayman-alkom:
In fact I am not sure about this, may need test

yurikuzn:
Yes. It changes every list in the system. Please, don't rely only on me testing PR.

yurikuzn:
https://github.com/espocrm/espocrm/commit/0721bd5c27550bd8863696edba5a235e5e2bfd30

ayman-alkom:
As I see you only made like a code clean & format, so the problem still exist.
I don't want to make this change privately, and instead I can improve the PR so the list will have another parameter `forceHideTotalCount` so that can be overridden in the dashlets

yurikuzn:
There a multiple places other than dashlets. `forceHideTotalCount` is not fine for us.

PR #1250, opened 2019-02-21
sum() aggregate fix

barwi:


yurikuzn:
Thanks. 

Actually it breaks the application since the interface is needed to be fixed too. I will do it by myself.

PR #1258, opened 2019-02-26
pdf template custom page size.

ayman-alkom:
the same pr but to different branch

### DIFF
--- a/application/Espo/Resources/i18n/en_US/Template.json
+++ b/application/Espo/Resources/i18n/en_US/Template.json
@@ -14,6 +14,8 @@
         "variables": "Available Placeholders",
         "pageOrientation": "Page Orientation",
         "pageFormat": "Paper Format",
+        "pageWidth": "Page Width (cm)",
+        "pageHeight": "Page Height (cm)",
         "fontFace": "Font"
     },
     "links": {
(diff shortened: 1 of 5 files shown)


yurikuzn:
Thanks.

yurikuzn:
Do we need pageOrientation field for Custom size? I think we can hide it.

ayman-alkom:
the current behavior respect the value of pageOrientation so that if you replaced between the pageWidth & pageHeight value you will get the same result
if we removed the pageOrientation then the value orientation must be calculated (may read only better than hidden)

yurikuzn:
Should we support decimal numbers (float) for size?

yurikuzn:
I think yes. TCPDF must support it.

yurikuzn:
Also I need to make size fields required if custom is selected.

ayman-alkom:
you are right, I can do them

ayman-alkom:
also the width must be larger than the (left + right margins + 1), the same for height

ayman-alkom:
> Should we support decimal numbers (float) for size?
we can also keep the integer and refer to it as mm instead of cm, the we will update the pdf service to remove `* 10` as well

yurikuzn:
https://github.com/espocrm/espocrm/commit/97a572cf966408ca80297935e71cab107e3cf573

PR #1264, opened 2019-03-06
Update Footer Year

gundogduyakici:
We don't have to change every year.


gundogduyakici:
So, Do you think the file changes are correct?

yurikuzn:
It's not fair to show the current year here.

PR #1267, opened 2019-03-08
dynamic logic: support file and image fields

ayman-alkom:
it is tested & works fine, only you may have another way to perform it


yurikuzn:
https://github.com/espocrm/espocrm/commit/3dcba630fe798951ad2acc9f98266982b1a22c1d

PR #1294, opened 2019-04-15
Fix getLevel method in acl manager

yunga91:
Hello.
If set disabled entity permission the ACL table would contain boolean value `false` for this entity. So in method getLevel we should check if `this.data.table[scope]` is object for using `in` operator then. In other case it cause error `Uncaught TypeError: Cannot use 'in' operator to search for 'read' in false`


yurikuzn:
Hi,

Fixes should be done in hotfix branch. Thanks.

beppe9000:
What is the expected release date for the hotfix ?

PR #1311, opened 2019-05-13
Show localized labels in header instead of fields names

vlasix:
In CSV export header, now we show localized labels instead of field names.


yurikuzn:
It uses system attributes names on purpose. For export/import.

vlasix:
Hi @yurikuzn , how would you like it if I added an option on the export screen that lets me choose whether I want labels or field names?

yurikuzn:
Hi, I'm not sure whether it's needed. Moreover I'm not sure about compatibility with our advanced pack extension.

CSV exports attributes, not fields. Translation for some attributes will be missed.

PR #1320, opened 2019-05-29
theme manager: fix stylesheet default path

ayman-alkom:


yurikuzn:
Should be client/css/espo/...

PR #1322, opened 2019-05-30
fix: override login view from config

ayman-alkom:
It is a nice feature to be able to override the login view, and it is already has been implemented,
but the problem is that metadata will not initialized before the login,
I hope to accept this change

### DIFF
--- a/client/src/controllers/base.js
+++ b/client/src/controllers/base.js
@@ -31,7 +31,7 @@ define('controllers/base', 'controller', function (Dep) {
     return Dep.extend({
 
         login: function () {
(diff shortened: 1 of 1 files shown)


ayman-alkom:
I hope you having a very nice holiday :)

yurikuzn:
Sorry, no.

ayman-alkom:
sorry to know that, and sorry because you not clearing why not

yurikuzn:
https://github.com/espocrm/espocrm/commit/4dc08dd37c12b8db6ae79cb89acfc9c1ffd0a09d

ayman-alkom:
thanks

PR #1324, opened 2019-06-04
Rendering of Multi-Enum fields on export to XLSX

Stas-Ghost:
Now Multi-Enum fields are exported to XLSX as _["Value1","Value2"]_
My goal is to restore behavior from version 4.x, remove brackets and quotes – _Value1,Value2_


yurikuzn:
Thanks. https://github.com/espocrm/espocrm/commit/f9223087df7a42cc7ac67f080b8167d0460ffae6

PR #1344, opened 2019-07-10
 more dynamic detail view and layout manager based on bootstrap table style

keredra:
Hi,

i did some rework on layout manager and detail view to support a more dynamic styling:

- switch panel mode between row mode and column mode
- allow more than two fields per row/column
- set available field area at fixed Position
- added address field for google maps routing service

TBD: it would be nice to be able to customise the column/row width.
Maybe i do it later when i have time for that.

Best regards
keredra


bandtank:
I love the dynamic layout. That would be so helpful.

hturkoz:
Hello keredra, thanks for this pull.. 
but for Google Maps Direction, i see you get "origin" in config file.. do you think is better "geolocate" directly the user ?
Best regards

keredra:
Hello hturkoz,
it is not so easy to geolocate devices without gps-function. Maybe it could be added as an option for mobile devices.

Best regards

hturkoz:
Thanks for response keredra, 
can you look here : https://developers.google.com/maps/documentation/javascript/examples/map-geolocation
it's juste that for me : 
 navigator.geolocation.getCurrentPosition(function(position) {
            var pos = {
              lat: position.coords.latitude,
              lng: position.coords.longitude
            };

so if user refuse allow "position" in navigator, you can then take the value origin in config file.
sorry, I have not skill for many think.. like this for sample ;)
Best Regards

keredra:
Some feedback from a developer would be nice.
Do not want to spend more time on this if they dont want to use it.

PR #1396, opened 2019-08-16
Feature: show/hide fields on record views in edit mode with dynamic conditions

superewald:
This will add the dynamic condition type `onEdit` which can be used to hide panels/fields in record detail views when in detail mode. We used it to add important notes (read-only text fields with default values) to the detail view while in edit mode.
It's just a little change but - at least for us - had a great improve in quality of workflow processing.

```json
"conditionGroup": [
    {
        "type": "onEdit"
    }
]
```

### DIFF
--- a/client/src/dynamic-logic.js
+++ b/client/src/dynamic-logic.js
@@ -31,6 +31,9 @@ define('dynamic-logic', [], function () {
     var DynamicLogic = function (defs, recordView) {
         this.defs = defs || {};
         this.recordView = recordView;
+        
+        recordView.on('after:set-edit-mode', function() { this.process() }.bind(this));
+        recordView.on('after:set-detail-mode', function() { this.process() }.bind(this));
 
         this.fieldTypeList = ['visible', 'required', 'readOnly'];
         this.panelTypeList = ['visible'];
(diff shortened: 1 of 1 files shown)


rodrigoscoelho:
Great

Enviado utilizando um Galaxy S9...

Em qui, 15 de ago de 2019 23:48, superewald <notifications@github.com>
escreveu:

ayman-alkom:
I suggest to have edit layout like detail layout, 
if no edit layout then the detail layout will be used as the default value,
I would like to hear @yurikuzn about these idea

superewald:
> I suggest to have edit layout like detail layout,
> if no edit layout then the detail layout will be used as the default value,
> I would like to hear @yurikuzn about these idea

I'm not sure if wether I get you right here. You mean enabling the condition type in the GUI (Layout Manager)  ?

ayman-alkom:
@superewald I suggested an additional idea, to have a new layout, just like the detail, detail small, list, list small.

so we can add new layouts named edit, edit small,

yurikuzn:
Having separate 'edit' layout is not on option.

hturkoz:
> > Having separate 'edit' layout is not on option.
> 
> Can you maybe investigate a bit further on why this isn't an option ?

hello, 
I think because "custom conditional logic" for view.  (layout)
so we need custom conditional logic for edit, for detail.
regards

yurikuzn:
Because detail and edit layouts must coincide. This was the concept from the very start of EspoCRM development.

And when you click edit, you are still on detail view. Only fields change their mode. The whole page is not re-rendered.

yurikuzn:
I'd use dynamic handler for hiding/showing fields.

Issue #1398, opened 2019-08-17
Duplicate Keys in Settings.json


karthik-uc:
https://github.com/espocrm/espocrm/blob/master/application/Espo/Resources/i18n/en_US/Settings.json#L94

b2CMode

yurikuzn:
https://github.com/espocrm/espocrm/commit/fe0dcc53427174ffdf59aa36fc8c5a96f757b271

PR #1399, opened 2019-08-17
#1398 duplicate keys removed and install/config.php ignored in git

theBuzzyCoder:
#1398


yurikuzn:
Please make sure that PRs don't contain any changes not related to an issue. This PR has the modification of gitignore file. What if I apply it being not cautious.

karthik-uc:
@yurikuzn Sorry about that. Will not do that again.

PR #1401, opened 2019-08-20
Feature: Add submenu with entity filters in navbar

superewald:
This will add the possibility to enable a submenu in the navbar with filters of the corresponding entity.

## Usage
- Add `navFilterList` to `clientDefs/{entity}.json`
```json
"navFilterList": [
  {
    "filter": "{filterName defined in filterList}",
    "icon": "{icon class eg: 'fas fa-address-card'}",
    "color": "{color of the icon in menu}
  }
]
```

## Example
*Expanded Navbar*
NavbarFilterSubmenuExpanded
*Minimized Navbar*
NavbarFilterSubmenuMinimized

This isn't the biggest change but it makes using the menu more semantic.
What do you guys think about this ?

> Note this is my current state of development, I will most likely refactor this to be more generic.

### DIFF
--- a/client/lib/handlebars.js
+++ b/client/lib/handlebars.js
@@ -1,6 +1,6 @@
 /*!
 
- handlebars v2.0.0
+ handlebars v3.0.3
 
 Copyright (C) 2011-2014 by Yehuda Katz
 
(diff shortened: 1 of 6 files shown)


yurikuzn:
Can be useful for some. Maybe would be better to have such an ability as an extension, . Not sure whether  we need to add this as an out-of-the-box feature. I have multiple ideas how to "improve" navbar but I'm not sure about any of them. I prefer not adding anything I'm not sure about its relevance, Once we add something we need support it forever.

cmediade:
I like! Would be a really good extension. Maybe OS like real estate.

Diese Email wurde vom Mobiltelefon gesendet.

Yuri Kuznetsov <notifications@github.com> schrieb am Di., 20. Aug. 2019,
20:32:

ayman-alkom:
it is not easy to customize navbar.js in extension

superewald:
> Can be useful for some. Maybe would be better to have such an ability as an extension. Not sure whether we need to add this as an out-of-the-box feature. I have multiple ideas how to "improve" navbar but I'm not sure about any of them. I prefer not adding anything I'm not sure about its relevance, Once we add something we need support it forever.

I totaly agree with you. The navigation is depending on the use case, so it would be nice to make customization optional. I would've also made this as an extension but to be honest I didn't found a proper way to implement it per extension.

> it is not easy to customize navbar.js in extension

That's exactly my problem. If @yurikuzn may could give me a hint in the right direction (how I can customize navbar.js and navbar.tpl per extension) then I'll for sure make this an extension.

yurikuzn:
I added the ability to define custom navbar view https://github.com/espocrm/espocrm/commit/1ef60781a028741e12fba0aedb86f1d117a83917

superewald:
> I added the ability to define custom navbar view [1ef6078](https://github.com/espocrm/espocrm/commit/1ef60781a028741e12fba0aedb86f1d117a83917)

Nice, thank you for that :)

PR #1467, opened 2019-10-11
Covered all function's test scenarios for Espo/Core/Utils/Config.php

karthik-uc:


karthik-uc:
If you're closing this pull request let me know the reason behind it. Any improvement suggestions are always appreciated.

tests/unit/Espo/Core/Utils/ConfigTest.php
```diff
@@ -150,8 +162,9 @@ public function testRemove()
         $this->assertTrue($this->object->save());
 
         $this->assertTrue($this->object->remove($optKey));
+        $this->assertTrue($this->object->save());
 
-        $this->assertNull($this->object->get($optKey));
+        $this->assertNull($this->object->remove($optKey));
```
yurikuzn: What's the reason of this ?

karthik-uc: This was to test remove when the $optKey is not present. That part of the code was not covered which i saw in code coverage report. If the $optKey was present then remove would return True and the test would fail.

karthik-uc: https://github.com/espocrm/espocrm/blob/master/application/Espo/Core/Utils/Config.php#L171

PR #1586, opened 2020-01-24
Try class in /custom before aborting command.php

### LINKED ISSUE #1585: Loading commands from custom path
If the class name is not standard then metadata should be used to specify it. 

But for classes following the standard custom namespace convention the class name can be inferred from metadata.

<hr>

Example implementation of custom command to echo all parameters, flags and options below.

```PHP
<?php

    namespace Espo\Custom\Core\Console\Commands;

    class DumpOptions extends \Espo\Core\Console\Commands\Base
    {
        public function run($options, $flagList, $argumentList) {
            echo "\n>> OPTIONS: "; print_r($options);
            echo "\n>> FLAGS: "; print_r($flagList);
            echo "\n>> ARGUMENTS: "; print_r($argumentList);
            echo "\nEnd of run() method.\n\n";
        }
    }
```

Example usage:

1. Put the file in `custom/Espo/Custom/Core/Console/Commands/DumpOptions.php`.
2. Run in ssh or batch like this

```text
$ php command.php dump-options -flag --option=value argument1 argument2

>> OPTIONS: Array
(
    [option] => value
)

>> FLAGS: Array
(
    [0] => flag
)

>> ARGUMENTS: Array
(
    [0] => argument1
    [1] => argument2
)

End of run() method.

```

beppe9000:
Resolves #1585

### DIFF
--- a/application/Espo/Core/Console/CommandManager.php
+++ b/application/Espo/Core/Console/CommandManager.php
@@ -67,6 +67,7 @@ class CommandManager
 
         $className = '\\Espo\\Core\\Console\\Commands\\' . $command;
         $className = $this->container->get('metadata')->get(['app', 'consoleCommands', $command, 'className'], $className);
+        if (!class_exists($className)) $className = '\\Espo\\Custom\\Core\\Console\\Commands\\' . $command;
         if (!class_exists($className)) {
             $msg = "Command '{$command}' does not exist.";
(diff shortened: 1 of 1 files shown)


beppe9000:
this is probably better to be put in a custom `ConsoleCommandManager` loader

PR #1600, opened 2020-02-07
relationship panel recordsPerPageSmall

eymen-elkum:
Hi @yurikuzn 

I found this needed, what do u think ?


yurikuzn:
It's fine. Name `recordsPerPage` would be better.

eymen-elkum:
will update it soon

PR #1619, opened 2020-02-19
[Templates] Fix es_ES traslation of Base entity

djgera:


yurikuzn:
This has be applied to hotfix branch.

PR #1624, opened 2020-02-22
improved CommandManager customization surface

beppe9000:
is this ok? please provide feedback

### DIFF
--- a/application/Espo/Core/Console/CommandManager.php
+++ b/application/Espo/Core/Console/CommandManager.php
@@ -32,6 +32,11 @@ namespace Espo\Core\Console;
 class CommandManager
 {
     private $container;
+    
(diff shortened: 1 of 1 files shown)


yurikuzn:
I see no reason for this changes. CommandManager is not supposed to be extended.

yurikuzn:
I can afford changes in such critical places. This is not covered with tests well. Risk to break upgrading for example, that will be a disaster.

beppe9000:
I shall revisit this once command handling is more mature

yurikuzn:
It's already fine. If you want something very very custom -- just create a separate php file.

beppe9000:
You mean by overriding the `consoleCommandManager` loader in `/custom`  and returning a different class?

PR #1687, opened 2020-04-27
allow custom view for portal navbar

eymen-elkum:
Hi @yurikuzn please check this, best regards.


yurikuzn:
I recommend to check in your custom navbar:

```
this.getUser().isPortal()
```

eymen-elkum:
You are right, but we need to include the custom script in the default portan-navbar view for better safe upgrade, as you may add more code to the portal-navbar view in the future.

I think if espocrm need different views for portal and main
then the extension must follow the same logic.

eymen-elkum:
thank you ...

yurikuzn:
Thanks for fix.

Issue #1715, opened 2020-05-22
./client/src/views/fields/person-name.js - this.model === 'edit' => this.mode?


hdijkema:
Could it be that there's a small code-mistake in:

- ./client/src/views/fields/person-name.js

?

```javascript
if (this.model === 'edit') {
   data.firstMaxLength = this.model.getFieldParam(this.firstField, 'maxLength');
   data.lastMaxLength = this.model.getFieldParam(this.lastField, 'maxLength');
   data.middleMaxLength = this.model.getFieldParam(this.middleField, 'maxLength');
}
```

I would think, `'this.model'` should be `'this.mode'`.

yurikuzn:
Thanks https://github.com/espocrm/espocrm/commit/c706ddc809e8ba870ad8ee8e6e2e24cbf3e0660c

PR #1716, opened 2020-05-22
fix typo

eymen-elkum:
fix https://github.com/espocrm/espocrm/issues/1715


yurikuzn:
I've already done it. Thanks.

PR #1717, opened 2020-05-22
update bootstrap datepicker library

eymen-elkum:
Some users complained about date picker in espocrm that not hold the input on enter key press.
I tried with the last version of the library and found it working perfectly.

To see the issue in action, just go to any date field (Task "Date Due" for example), try to write the date using keyboard (many users like using keyboards more than mouse), and press enter.

The expected behaviour is to close the datepicker and accept the entered date and apply the format.

What happen is the datepicker closed but the input lost, I noticed that this happens even if we clicked on the day cell using the mouse while the input filled manually.

Also it is good idea to update the lib after some years.

Best Regards.

### DIFF
--- a/client/lib/bootstrap-datepicker.js
+++ b/client/lib/bootstrap-datepicker.js
@@ -20,1652 +20,2036 @@
  * limitations under the License.
  * ========================================================= */
 
-(function($, undefined){
-
-	var $window = $(window);
-
-	function UTCDate(){
-		return new Date(Date.UTC.apply(Date, arguments));
-	}
-	function UTCToday(){
-		var today = new Date();
-		return UTCDate(today.getFullYear(), today.getMonth(), today.getDate());
-	}
-	function alias(method){
-		return function(){
-			return this[method].apply(this, arguments);
-		};
-	}
-
-	var DateArray = (function(){
-		var extras = {
-			get: function(i){
-				return this.slice(i)[0];
-			},
-			contains: function(d){
-				// Array.indexOf is not cross-browser;
-				// $.inArray doesn't work with Dates
-				var val = d && d.valueOf();
-				for (var i=0, l=this.length; i < l; i++)
-					if (this[i].valueOf() === val)
-						return i;
-				return -1;
-			},
-			remove: function(i){
-				this.splice(i,1);
-			},
-			replace: function(new_array){
-				if (!new_array)
-					return;
-				if (!$.isArray(new_array))
-					new_array = [new_array];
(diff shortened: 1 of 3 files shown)


eymen-elkum:
I kept the license header comments and replaced all the codes from the original source:

https://raw.githubusercontent.com/uxsolutions/bootstrap-datepicker/master/dist/js/bootstrap-datepicker.js

yurikuzn:
It needs to be tested before applying. Usually updates bring surprises.

eymen-elkum:
I made many tests, and will make more

I agree with you about updates surprises, but the issue needs this update 

I will confirm here when I finish my tests.

yurikuzn:
I will test too.

yurikuzn:
It's too much changes for hotfix. Chances that something will be broken are high.

yurikuzn:
It worked since 2013 the way how it works now. No need to hurry to apply changes in hotfix. Hotfix can be released in any moment as a critical bug occur. I'd rather move it to following hotfix (5.9.3 or later) or master.

eymen-elkum:
I understand your worry about the changes but I think no need to worry as we update from popular library, they already noticed the issues and fixed.

I also searched for the using of datepicker and only found one use in [fields/date.js](https://github.com/espocrm/espocrm/blob/e4237abdbf6605130f9d32978f8f42c534742ee4/client/src/views/fields/date.js#L159)

I compared all options with the docs of the bootstrap-datepicker and all are correct, not breaked.

I tested the edit & search modes everything seems perfect.

One test remaining is to build the espocrm using grunt, then repeat the test on compiled js & css

eymen-elkum:
> It worked since 2013 the way how it works now. No need to hurry to apply changes in hotfix. Hotfix can be released in any moment as a critical bug occurs. I'd rather move it to the following hotfix (5.9.3 or later) or master.

👍 I agree

yurikuzn:
It's not about whether lib is reliable since it's popular. We often have issues when apply such UI libs. There are too many places where problems may occur.

eymen-elkum:
> It's not about whether lib is reliable since it's popular. We often have issues when apply such UI libs. There are too many places where problems may occur.

As I said we have only one use over all the crm, not sure for the advanced back, 

Any way I think will not take more than one hour from to make sure that everything is okay.

I now looking for a way to fix this issue for my client without waiting for the update

I suggest you to ask about this issue in the forum, many clients will sends valuable feedbacks.

I will close this PR and make new one to the master so you can merge any time.

Best Regards

PR #1769, opened 2020-08-07
Fix date wrong value in date detail/edit view when non UTC timezone is selected

kaminarifox:
Detail view image

Edit view image


yurikuzn:
https://github.com/espocrm/espocrm/issues/1770

PR #1782, opened 2020-09-08
edit-link-modal: keep elements positions

eymen-elkum:
Hi @yurikuzn 

I hope this is helpful, because in recent versions the edit links modal elements not keeping the same position, so that may cause confusion for the user

Screen Shot 2020-09-08 at 5 07 37 PM
Screen Shot 2020-09-08 at 5 08 44 PM

Best Regards


yurikuzn:
Thanks.

Here's my fix: https://github.com/espocrm/espocrm/commit/72fd9184f4e7f43f6a88c51439b02774ec19f1bf.

PR #1786, opened 2020-09-15
minor code improve

eymen-elkum:
the old version of createView function was accepting the 4th param as context, but now it is for wait, so we can improve the code using that parameter.

### DIFF
--- a/client/src/views/stream.js
+++ b/client/src/views/stream.js
@@ -62,7 +62,6 @@ define('views/stream', 'view', function (Dep) {
         setup: function () {
             this.filter = this.options.filter || this.filter;
 
-            this.wait(true);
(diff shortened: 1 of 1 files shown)


yurikuzn:
This fix won't work properly. this.getModelFactory().create is asynchronous.

eymen-elkum:
yess :) you are right!

yurikuzn:
For future it's better to use promises:
```js

        this.wait(
            this.getModelFactory().create(entityType)
            .then(
                function (model) {
                    this.model = model;
                }.bind(this)
            )
            .then(
                function () {
                    this.createView( ... );
                }.bind(this)
            )
        );

```

eymen-elkum:
Very nice, thank you!

PR #1797, opened 2020-10-01
Fields ACL: additional way to define it directly from field manager

eymen-elkum:
Hi @yurikuzn 

This is about an additional way to control fields access directly from the entity defs, in additions to the existing `entityACL` metadata.

This is more easy for developers to define the ACL for new type fields or existing fields, and even can be set from field manager in the future.

Example:
[This ACL here](https://github.com/espocrm/espocrm/blob/master/application/Espo/Resources/metadata/entityAcl/User.json#L4):
Screen Shot 2020-10-01 at 2 48 47 PM

can be set as following from the entityDefs:
Screen Shot 2020-10-01 at 2 45 37 PM

If you liked the idea and stuck with minor code issues would like to tell me to update ;)

### DIFF
--- a/application/Espo/Core/Acl/GlobalRestricton.php
+++ b/application/Espo/Core/Acl/GlobalRestricton.php
@@ -144,6 +144,22 @@ class GlobalRestricton
                 $scopeData->fields->$type = $resultFieldList;
                 $scopeData->attributes->$type = $resultAttributeList;
             }
+
+            foreach ($fieldList as $field) {
+                $fieldAcl = $this->metadata->get(['entityDefs', $scope, 'fields', $field, 'acl']) ?? [];
+
+                foreach ($this->fieldTypeList as $type) {
+                    if ($fieldAcl[$type] === true) {
+                        $isNotEmpty = true;
+                        $scopeData->fields->$type[] = $field;
+                        $fieldAttributeList = $this->fieldUtil->getAttributeList($scope, $field);
+                        foreach ($fieldAttributeList as $attribute) {
+                            $scopeData->attributes->$type[] = $attribute;
+                        }
+                    }
+                }
+            }
+
             foreach ($this->linkTypeList as $type) {
                 $resultLinkList = [];
 


yurikuzn:
Hi,

I don't like an idea having the same parameter in multiple places.

eymen-elkum:
I am too, 

In the future we may can drop the antityACL and replace with using the entityFeds directly.

EntityDefs is main place where we define the fields and their params, will be great if we also can define the accessibility, you know how much manipulating with entityDefs is easier than entityACL.

I think that entityACL is made for only manually setting.

Please give this idea one more thought.

yurikuzn:
I don't see how manipulating entityDefs is easiear than entityAcl. I really don't see  problem that acl parameters is placed separated. I like them being separate from other parameters.

yurikuzn:
I would like to stick with the way how it is.

eymen-elkum:
To explain my idea more:

Think about the field manager having that dropdown params for the accessibility, how much will be great!!

Screen Shot 2020-10-01 at 3 08 38 PM

yurikuzn:
Ok. I seems understand your point now. It would be easier to have this parameters in Field Manager.

eymen-elkum:
So happy to hear that

yurikuzn:
Moving to entityDefs is not an option. It's too much of backward compatibility problems.

Handling in backend so that it stores/fetch to/from entityAcl is not a big deal.

yurikuzn:
entityAcl was not supposed to be customizable. Field Level Security is supposed to handle an access. entityAcl was for intended for system field first.

eymen-elkum:
Yes I know, but this PR bring that great work to be used easily from entityDefs, and it is totally compatible with any recent version of espocrm

eymen-elkum:
Anyway, 

In future, if you decided to move to entity defs we can design the afterUpgrade function to do the magic, I believe it is less than 20 line of php code :)

yurikuzn:
It's not as simple as it may seem. I spend days investigating for such small things that may seem trivial.

eymen-elkum:
> It's not as simple as it may seem. I spend days investigating for such small things that may seem trivial.

I think about reading the entityAcl and write in the properly places in the entityDefs

As I said this is another topic, and I can help with that if you you allow, 

For now I really will appreciate this PR to be merged, because I searched a lot and not found better than this to do the job.

yurikuzn:
I'm too busy with 6.0.0. I can't digress on anything else now. I can not merge it. Sorry.

eymen-elkum:
Okay no problem,
I will merge manually for the time being on my projects, and you may merge later when you have the time, 

thank you so much!

yurikuzn:
There's already 'readOnly' parameter in entityDefs that only affects UI. After your change, you will have this parameter affecting access control. I tested it once, making 'readOnly' affecting be honored by an access control and had some incompatibility issues.

I'd recommend you customizing FieldManager.

yurikuzn:
> In future, if you decided to move to entity defs we can design the afterUpgrade function to do the magic, I believe it is less than 20 line of php code :)

yurikuzn:
> In future, if you decided to move to entity defs we can design the afterUpgrade function to do the magic, I believe it is less than 20 line of php code :)

This is not the way how things work. We have also backward compatibility burden.

Please consider that even a minor change in core functionality may bring problems for us.

eymen-elkum:
First of all, many thanks to give it some time while you are busy with the version 6.0.0

Yes, `readOnly` is shared between the two engines, but I am not sure if this will cause a problem, I thought that this will be enhancement rather than make a problem as the field will be backend write protected.

For now when we set the `readOnly: true` in entityDefs this will not really protect the field from editing using API.

Any way, we have a good option, it is adopting the following approach:

Screen Shot 2020-10-01 at 8 35 30 PM

What do you think?

If you liked the idea please tell me to update the commit, I will be happy to complete this PR.

eymen-elkum:
Hi @yurikuzn ,

I updated the PR so the acl defs will be hold inside "acl" param, this will protect against any potential conflicts like `readonly`

I have used this patch in many production projects and there is no issues at all, you can also test from your side.

PR #1804, opened 2020-10-07
fix: Date field 'before' and 'after' validations 

kaminarifox:


yurikuzn:
It breaks an existing functionality.

kaminarifox:
> It breaks an existing functionality.

At this moment before/after validators not working at all. Is there a proper way to fix it?

yurikuzn:
They are working. Try creating meeting with date end before date start and you will get an invalid message.

kaminarifox:
Meeting use the _datetime-optional_ field, but i talk about the _date_ field in custom entities

yurikuzn:
It works for date field as well.

kaminarifox:
In the validator `otherValue` always undefined, because `field` is just date string

https://github.com/espocrm/espocrm/blob/d02529ee86b84c150fec6a86c4ac5be2d90bc611/client/src/views/fields/date.js#L329

yurikuzn:
I think you misunderstand how this validation supposed to work.

kaminarifox:
Oh, now i see. Sorry for wasted time.

Have you thought about adding a validator depended on date string or making current validators more universal?

yurikuzn:
Yes. I have this in my TODO list.

PR #1809, opened 2020-10-13
admin panel search backward compatibility

eymen-elkum:
Old extensions versions (including sales & advanced) not have the keywords attribute and this cause a bug in the console

Regards


yurikuzn:
I believe it's already fixed by https://github.com/espocrm/espocrm/commit/7349f37786bed2b35d1bf08026dcf133234d480a

eymen-elkum:
Yes, you are right.

My master is a bit old

PR #1864, opened 2020-12-13
array field: improve select modal

eymen-elkum:
Hi @yurikuzn 

I hope you like this improvement, it is about the array field item select modal.

As you know if we want to add several items then we have to open the modal several times, and it is closing automatically after every selection which may not good for UX.

In this PR I made it close automatically only when all items selected,

### DIFF
--- a/client/src/views/fields/array.js
+++ b/client/src/views/fields/array.js
@@ -577,7 +577,7 @@ define('views/fields/array', ['views/fields/base', 'lib!Selectize'], function (D
             return this.getSearchParamsData().type || 'anyOf';
         },
 
-        actionAddItem: function () {
+        getAddModalOptions: function () {
             var options = [];
 
             this.params.options.forEach(function (item) {
@@ -586,13 +586,24 @@ define('views/fields/array', ['views/fields/base', 'lib!Selectize'], function (D
                 }
             }, this);
 
-            this.createView('addModal', this.addItemModalView, {
+            return {
                 options: options,
                 translatedOptions: this.translatedOptions,
-            }, function (view) {
+            };
+        },
+
+        actionAddItem: function () {
+            this.createView('addModal', this.addItemModalView, this.getAddModalOptions(), function (view) {
                 view.render();
-                this.listenToOnce(view, 'add', function (item) {
+                view.once('add', function (item) {
                     this.addValue(item);
+                    view.close();
+                }.bind(this));
+
+                view.once('add-mass', function (items) {
+                    items.forEach(function (item) {
+                        this.addValue(item);
+                    }.bind(this));
                     view.close();
                 }.bind(this));
             }.bind(this));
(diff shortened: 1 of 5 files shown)


client/src/views/fields/array.js
```diff
@@ -82,9 +82,8 @@ define('views/fields/array', ['views/fields/base', 'lib!Selectize'], function (D
                     translatedOptions: this.translatedOptions
                 }, function (view) {
                     view.render();
-                    this.listenToOnce(view, 'add', function (item) {
+                    view.on('add',function (item) {
```
eymen-elkum, line 85: I used the event directly from the view object, so after closing no more listining to the old modal, as we may open the modal several times.

yurikuzn:
Hi,

I think it's reasonable to have the possibility to add multiple items w/o model being closed. But I think it's not good if we remove the current behavior that requires 2 clicks to add one item. After this change it will require an additional click to close modal.

Maybe it would be better to have 2 way of adding items. The current one and checkboxes. I will think about it.

eymen-elkum:
Ok, what do you think about having an additional field on the field manager named for example "**Disable auto close modal**"

Another idea - but not the best - is to have additional input next to the "add" button that enable / disable this mode on the fly.

Best Regards

If you like someone help improving with you I will happy to perform any of these ideas or any idea you suggest!

eymen-elkum:
Hi @yurikuzn ,

I have updated the pull request, adding checkboxes for better UX regarding your valuable notes:

Screen Shot 2020-12-26 at 1 50 28 PM

yurikuzn:
Nice. Could you change the branch to `master`?

Why buttons are placed in the bottom on your screenshot? Is it a customized theme?

eymen-elkum:
> Nice. Could you change the branch to master?

For sure!

> Is it a customized theme?
Yes

eymen-elkum:
I will resolve the conflects now ...

eymen-elkum:
I have updated the PR to fit the master last changes, 

I have extracted a new function anmed `getAddModalOptions` to help the `tab-list` view to ocerride the options of the addModal without the need to overrite the whole enclosed function, **this fixed an error** [here](https://github.com/espocrm/espocrm/blob/76243dd64d6445082c3194b7dd93d9c7eb57b597/client/src/views/settings/fields/tab-list.js#L98) where the array is object.

yurikuzn:
I'd use another event `add-mass` for adding multiple items at once. I don't like having the same event for one or multiple items.

eymen-elkum:
Good idea!
I got you and going to make the changes now...

eymen-elkum:
Hi Yuri, I hope everything is okay regarding this PR, I have discarded the code formatting for better focus on the real changes.

I can push the code formatting if you wish.

yurikuzn:
Please see my comment regarding checkedList parameter initialization. The way how it is initialized will cause problems.

yurikuzn:
Good. Thanks. I've made some fixes: https://github.com/espocrm/espocrm/commit/b6a6cf052b0e9be77f4a86525511767e50b8e334

PR #1882, opened 2021-01-05
link field keep name value in search mode

eymen-elkum:
Hi @yurikuzn ,

As I love using keyboard more than the mouse, I noticed that when we press enter, name-value is missing, so the user will not see the saved selected entity name any more.

This issue will not happpen if I clicked on the result item using the mouse, or if I used the select modal.

Screen Shot 2021-01-05 at 9 21 11 PM


yurikuzn:
Thanks

PR #1885, opened 2021-01-17
Apply `createDisabled ` in relationship panels

eymen-elkum:
Hi @yurikuzn 

This is for applying the nice `createDisabled` from clientDefs to relationship panels.

I am using this small change in production and would be happy to see it merged in the next version.

Best Regards

### DIFF
--- a/client/src/views/record/panels/relationship.js
+++ b/client/src/views/record/panels/relationship.js
@@ -64,7 +64,7 @@ define('views/record/panels/relationship', ['views/record/panels/bottom', 'searc
             var url = this.url = this.url || this.model.name + '/' + this.model.id + '/' + this.link;
 
             if (!('create' in this.defs)) {
-                this.defs.create = true;
+                this.defs.create = !this.getMetadata().get(['clientDefs', this.scope, 'createDisabled']);
             }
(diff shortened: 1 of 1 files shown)


yurikuzn:
Hi,

You should use`createdDisabled` in relationship defs. The main `createDisabled` has nothing to do with relationship panels.

eymen-elkum:
Thanks for the replay, but I am not satisfied
If you are interested I can logically explain the necessity of this change
The change is not so important, however

yurikuzn:
We need to have the ability to disable 'create' on the list view but be able to create related records.

eymen-elkum:
That makes sense, I agree!

But this PR doesn't affect that, it is all about making the `createDisabled` default value on relationship can be set from the main `createDisabled` in the clientDefs, **keeping** the possibility of overriding from the reationship defs.

To me it is better to define this value one time with the ability to override it on relationship panels rather than having to pass it for all relationship panels.

yurikuzn:
InEspo we follow the rule that `false` should never override `true`. It's better off not changing this here I think. Not a big deal to add the parameter to each relation panel.

PR #1892, opened 2021-01-21
htmlizer `trTag` avoid breakline

eymen-elkum:
Hope this is useful PR.

links:

https://sourceforge.net/p/tcpdf/discussion/435311/thread/ff6879d2/

### DIFF
--- a/application/Espo/Core/Htmlizer/Htmlizer.php
+++ b/application/Espo/Core/Htmlizer/Htmlizer.php
@@ -525,7 +525,7 @@ class Htmlizer
                 $context = $args[count($args) - 1];
 
                 return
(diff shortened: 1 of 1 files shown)


yurikuzn:
After I moved all TCPDF logic behind and interface, having TCPDF specific parameters before that interface is not good.

eymen-elkum:
Is there any woraround to make this work for TCPDF?

yurikuzn:
I'm not sure how would be better. 

Does the browser strip this `nobr` attribute when you put it in wysiwyg?

rabiibrahimi:
This is a very good PR. i have faced issues with this and would be great to solve this out. 
Thanks

PR #1906, opened 2021-02-02
Patch 15

eymen-elkum:
This fixed a bug that happens when the whole id chars are digits, 

The issue is as following:
- check "check all" checkbox
- uncheck a record that has "all-digits" id.
- mas actions will not exclode the unchecked id!!

My client was imported the data from anpther system,


yurikuzn:
We should have used `attr` method instead of `data`. I applied the fix to the hotfix branch.

https://github.com/espocrm/espocrm/commit/7320bf68e5dbe3abba3e2f81a9bedc3df032ee0f

eymen-elkum:
excelent!

PR #1949, opened 2021-03-23
integration fields view param

eymen-elkum:
In the current case not possible to define custom view for integration fields, hope this patch is ok


yurikuzn:
It's OK.

I prefer having more clean readable code like:

```
var viewName = (params || {}).view || this.getFieldManager().getViewName(type);

this.createView(name, viewName , {
```

Or to be consistent with codebase:

```
var viewName = this.model.getFieldParam(name, 'view') || this.getFieldManager().getViewName(type);
```

PR #1951, opened 2021-03-23
validation: small fix

eymen-elkum:
I hope this PR is useful,

`isNaN(undefined); // true`

ref: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/isNaN

### DIFF
--- a/client/src/views/fields/float.js
+++ b/client/src/views/fields/float.js
@@ -72,7 +72,7 @@ Espo.define('views/fields/float', 'views/fields/int', function (Dep) {
 
         validateFloat: function () {
             var value = this.model.get(this.name);
-            if (isNaN(value)) {
+            if (typeof value !== 'undefined' && isNaN(value)) {
                 var msg = this.translate('fieldShouldBeFloat', 'messages').replace('{field}', this.getLabelText());
                 this.showValidationMessage(msg);
                 return true;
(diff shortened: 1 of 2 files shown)


yurikuzn:
Looks wrong to me. Shouldn't it be `typeof value !== 'undefined'` ?

I prefer having single quotes for short strings. `'undifined'`.

yurikuzn:
I also now sure it's possible to have a situation that value is fetched as `undefined`. It's supposed to be `null` when empty https://github.com/espocrm/espocrm/blob/6.1.4/client/src/views/fields/int.js#L241.

eymen-elkum:
> Looks wrong to me. Shouldn't it be `typeof value !== 'undefined'` ?

You are right :)

eymen-elkum:
I faced this issue on currency field was raising the validation error even if empty and not required

yurikuzn:
Could not reproduce. Maybe you had whitespace?

eymen-elkum:
It was on custom field rendering not in standard record view

yurikuzn:
It would be better to modify parse or fetch method in the custom field.

eymen-elkum:
it is not custom field, but custom field wrapper rather than standard record view, I think this issue occured because of model inisialisation but while testing I found that checking if the value is undefined is better for future for other cases.

I didnot made further checking because I thought that this function must already check for undefined possibility.

eymen-elkum:
Ok, it seems I have to call `view.fetchToModel();` ,
this will set `null` to the fields that not toutched yet.

PR #1955, opened 2021-04-02
eventAssignedUserIsAttendeeDisabled on frontend

eymen-elkum:
Hi @yurikuzn 

This config key `eventAssignedUserIsAttendeeDisabled` is helpful ond applied well in the background, but not applied on the frontend, this time I pushed some whitespaces as well.

When some set this key to `true` it is not expected to see the assigned user on the scheduler panel


yurikuzn:
Thanks. I applied the fix to hotfix: https://github.com/espocrm/espocrm/commit/84a3688525b5d1806cf9982459a179d65408895c

PR #1976, opened 2021-04-26
New date display format

SinyaWeo:
This date format with spaces is correct one in Czech language.

### DIFF
--- a/application/Espo/Resources/metadata/entityDefs/Settings.json
+++ b/application/Espo/Resources/metadata/entityDefs/Settings.json
@@ -28,7 +28,7 @@
         },
         "dateFormat": {
             "type": "enum",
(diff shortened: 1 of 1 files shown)


yurikuzn:
I'm not sure whether everything will work with this format added. I'm concerned that we might rely on a string length somewhere in codebase to distinguish date and date-time strings. I need to investigate.

SinyaWeo:
Maybe change will be also needed in Portal.json file.

If I can help you with anything, just let me know. Thank you.

yurikuzn:
I refactored so that all date formats are listed only in one place metadata > app > dateTime.

Added the new format to the next release https://github.com/espocrm/espocrm/issues/1978.

SinyaWeo:
That is a great news and great work you've done. Thank you so much.

PR #1998, opened 2021-05-19
config param emailAutoReplaySuppressPeriod

eymen-elkum:
Hi Yuri,

This is following to the [forum thread](https://forum.espocrm.com/forum/general/70910-case-auto-reply-only-works-once-per-email-address#post70950).

There one more change maybe needed, adding more conditions like the email name, but it is outside this PR


yurikuzn:
Could you rename it to `emailAutoReplaySuppressPeriod`.

eymen-elkum:
renamed

yurikuzn:
Thanks

PR #2012, opened 2021-06-02
fix labels vertical gaps

eymen-elkum:
## After
Screen Shot 2021-06-02 at 18 16 16

## Before
Screen Shot 2021-06-02 at 18 16 43


yurikuzn:
Thanks. Required an additional fix: https://github.com/espocrm/espocrm/commit/01c0fa956366abbfce275ccf530868a40064b944

Issue #2009, opened 2021-06-01
loadAdditionalFields does not work for email templates


boydle:
Issue with email template merge fields. Fields that are populated using loadAdditionalFields do not have a value when merged in to an email template.

yurikuzn:
Please explain more clear and detail. Steps to reproduce would be helpful.

boydle:
I have built an example of the problem and attached it:

I have created a new entity "EmailTestMyth" I have added the following code to the service:
    public function loadAdditionalFields(Entity $entity)
    {
        parent::loadAdditionalFields($entity);
        $entity->set('firstName', 'FNAME');
        $entity->set('lastName', 'LNAME');
    }  

I added the  fields firstName and lastName to the detail layout and the values set in the service are shown on the detail view.

I created an email template based on the new entity with the following body:

*************
Name: {EmailTestMyth.name},
FName: {EmailTestMyth.firstName}
LName: {EmailTestMyth.lastName}
*************

I have added "EmailTestMyth" as an email record parent.

If i use the email template and pick a "EmailTestMyth" record the name field is populated but the firstName and lastName fields are not set.
[custom.zip](https://github.com/espocrm/espocrm/files/6601921/custom.zip)

PR #2035, opened 2021-07-09
Fix issue #2009

HungNguyenPTN:
-Fix loadAdditionalFields does not work for email templates
-Similar to PDF template
This fix for issue [#2009 ](https://github.com/espocrm/espocrm/issues/2009)

### DIFF
--- a/application/Espo/Services/EmailTemplate.php
+++ b/application/Espo/Services/EmailTemplate.php
@@ -105,6 +105,8 @@ class EmailTemplate extends Record implements
         if (empty($params['parent'])) {
             if (!empty($params['parentId']) && !empty($params['parentType'])) {
                 $parent = $this->getEntityManager()->getEntity($params['parentType'], $params['parentId']);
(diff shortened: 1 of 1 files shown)


yurikuzn:
Hi,

Thanks for PR. But it has an issue. The code should be inserted inside the `if` block.

HungNguyenPTN:
Hi, i updated.
inside `if (parent)` right?
thank you for responed

yurikuzn:
Looks right.

boydle (commented):
Hi, I think the 2 new lines of code should have been added at line 121 not where they are. With this code I think the loadAdditionalFields could be missed when $params['parent'] is set.

HungNguyenPTN:
Hi, i think @boydle  comment is more correct. But i cant edit anymore. How can i fix this

yurikuzn:
I'd leave as it is. The place where 'parent' is set should care about populating all fields I think. The less object mutation inside a service the better I believe.

PR #2045, opened 2021-07-30
template wrapper name getter

eymen-elkum:
Hi,

I am working on new PDF engine, and I am almost finished (except for barcode), I just needed this core change,

### DIFF
--- a/application/Espo/Tools/Pdf/Template.php
+++ b/application/Espo/Tools/Pdf/Template.php
@@ -62,4 +62,6 @@ interface Template
     public function getPageWidth() : float;
 
     public function getPageHeight() : float;
+
+    public function getName() : string;
 }
--- a/application/Espo/Tools/Pdf/TemplateWrapper.php
+++ b/application/Espo/Tools/Pdf/TemplateWrapper.php
@@ -119,4 +119,9 @@ class TemplateWrapper implements Template
     {
         return $this->template->get('pageHeight') ?? 0.0;
     }
+
+    public function getName() : string
+    {
+        return $this->template->get('name');
+    }
 }


yurikuzn:
Hi,

Can you explain why internal template name is needed for PDF printing.

eymen-elkum:
Sure,

The engine I used is not Pdf, but HTML, so I am displaying the result as html page

I need the template name as title of the page.

This was exciting experience to me, this way I can use the full power of css to make complex reports for clients

Soon later I will share more information about this extension

eymen-elkum:
This is a screenshot:

Screen Shot 2021-07-30 at 4 59 39 PM

yurikuzn:
I think we need to add a new field to template for Title rather than use existing Name.

eymen-elkum:
The name field is already used for PDF file name, adding a new field may not needed, I followed the same convension
But still good idea, so the function will return the name if the title is empty

eymen-elkum:
sometimes clients would like to append the printed entity name to the pdf file name.

yurikuzn:
Added title: https://github.com/espocrm/espocrm/commit/a56acf553c2f8b28ec74dbef3b34b105ed055841

I will think about supporting placeholders in title later.

yurikuzn:
Are you implementing headless chrome for PDF printing?

eymen-elkum:
> Are you implementing headless chrome for PDF printing?
To be honest, some features not implemented yet like repeating header every page, I may make this extension open source, then community can help.

yurikuzn:
FYI. I added `getStream` method to Contents interface to be able to work with large files (not actual for Tcpdf).

eymen-elkum:
Thank you so much for the information, I will check it!

PR #2047, opened 2021-07-31
PDF title placeholders support

eymen-elkum:


yurikuzn:
I didn't want to use htmlizer here so I didn't set it aside for the future consideration.

yurikuzn:
FYI. I made changes to interfaces:
* https://github.com/espocrm/espocrm/blob/master/application/Espo/Tools/Pdf/EntityPrinter.php
* https://github.com/espocrm/espocrm/blob/master/application/Espo/Tools/Pdf/CollectionPrinter.php

As it's a major release it's an opportunity to improve existing APIs.

eymen-elkum:
We are excited to see version 7 alive 👍🏼

PR #2053, opened 2021-08-08
Change default timeZone from config

farhadlavaei:


yurikuzn:
This is a breaking core change that changes all concept we rely to. All internal dates are in UTC.

farhadlavaei:
When is set the TimeZone in setting, the created_at field dose not save correct date, But when i change this , it correct.

yurikuzn:
It's by design is set in UTC timezone.

PR #2264, opened 2022-03-29
fix config.php

eymen-elkum:
Fix:
Rebuild database fault: Error: Cannot use object of type stdClass as array


yurikuzn:
There were already changes: https://github.com/espocrm/espocrm/blob/master/application/Espo/Core/Utils/Config.php#L114

When fixing such core things like the config, some unit tests are needed.

eymen-elkum:
Yes, seems fixed already there, thanks!

PR #2279, opened 2022-04-18
Reset model changed attrs once fetched

eymen-elkum:
This will help any developer to correctly check if model changed or not.

Current behavior: any field not listed on list layout will marked as changed.

Expected behavior: changed attributes must be null or empty object when just fetched from the server

### DIFF
--- a/client/src/views/modals/edit.js
+++ b/client/src/views/modals/edit.js
@@ -129,6 +129,7 @@ define('views/modals/edit', 'views/modal', function (Dep) {
                     }
 
                     model.once('sync', () => {
+                        model.changed = {};
                         this.createRecordView(model);
                     });
 


yurikuzn:
If I remember correctly, changed should be used only within 'change' event. It's erased on each `set` call. I believe you misusing this property.

eymen-elkum:
If you agree with the expected behavior that mentioned  above will be great if you manage for it in different way, personally found it the best available way.

eymen-elkum:
I also think that Backbone.Model not handled the case of making partially fetching, that is why they don’t reset attributes, 
In other words, they consider any new input like if server side changes, but in our case we only care about client side changes.

yurikuzn:
You can try comparing with attributes of `sourceModel`.

yurikuzn:
Also, the record/edit view has the logic that helps to detect what was changed. All initial attributes are stored in the `attributes` property. Compare model attributes with them to know what was changed.

eymen-elkum:
Thank you so much for the suggestions, I already managed for this, just thought it is better to be on the used model level directly.

Issue #2334, opened 2022-06-03
Default path for client/modules points to client/custom/modules


AgentT30:
**Describe the bug**
The default path for client/modules points to client/custom/modules in the return statement.

**To Reproduce**
Steps to reproduce the behaviour:
1. Create a custom entrypoint in a custom module with $runscript like
`
$runScript = "
Espo.require('custom-module-name:controllers/method_name', function (Controller) {
    var controller = new Controller(app.baseController.params, app.getControllerInjection());
    controller.masterView = app.masterView;
    controller.doAction('method_name', ".json_encode($data).");
});
";
`
2. Try to load the entrypoint in the browser.
3. When the entrypoint is loaded in the browser an error 404 is thrown saying `/client/custom/modules/custom-module-name/src/controllers/controller-name.js 404 (Not Found)`

**Expected behaviour**
The template given to the entrypoint must be loaded in the browser. The `client/custom/modules` should not be appended to the clients of the non 'custom' module. 

Tried the same by modifying the `espo.min.js` by removing the `client/custom/modules` and keeping `client/modules` the entrypoint works as expected.

**EspoCRM version**
7.1.7

**Additional context**
PHP Version 7.4.29
Browsers: Chrome and Firefox

AgentT30:
https://github.com/espocrm/espocrm/pull/2335 - PR.

arkadywtf:
This is not a bug. See here: https://github.com/espocrm/espocrm/issues/2032

yurikuzn:
Sorry, but this fixes your problem but creates a problem for all others who have their modules in `custom/modules` dir.

PR #2335, opened 2022-06-03
Default path for client/modules points to client/custom/modules #2334 

AgentT30:


yurikuzn:
This breaks the applications. Modules are supposed to be located in `custom/modules` dir since the recent release unless there are internal.

PR #2368, opened 2022-07-07
Fulltext search by related field

mozkomor05:
Made it possible to use fulltext on related attributes. Example:

```json
{
    ...
    "collection": {
        "orderBy": "createdAt",
        "order": "desc",
        "textFilterFields": [
            "product.name"
        ],
        "fullTextSearch": true,
        "countDisabled": false
    },
}
```


application/Espo/Core/Select/Text/Applier.php
```diff
@@ -106,7 +106,7 @@ public function apply(QueryBuilder $queryBuilder, string $filter, FilterParams $
             array_filter(
                 $this->metadataProvider->getTextFilterAttributeList($this->entityType) ?? ['name'],
                 function ($field) use ($fullTextSearchFieldList) {
-                    return !in_array($field, $fullTextSearchFieldList);
+                    return in_array($field, $fullTextSearchFieldList);
```
yurikuzn, line 109: This seems to me wrong.

yurikuzn:
Would MySQL work with multiple full-text fields across two tables in a single MATCH?

application/Espo/Core/Select/Text/Applier.php
mozkomor05: > This seems to me wrong.

You are right. I misunderstood the purpose of `DefautFilter::applyAttribute` method.

mozkomor05:
> Would MySQL work with multiple full-text fields across two tables in a single MATCH?

No it wouldn't, since `MATCH` needs index on all of the arguments in the same table. I'll work on the fix today.

yurikuzn:
I think supporting full-text search across multiple tables requires another design. The current design doesn't suit it. Trying it make it to suit will increase the complexity (it's quite complex already) of the logic. Multiple issue may occur.

yurikuzn:
From the first glance it increased complexity pretty much. I'm extremely busy at the moment, can't review. My idea was to not to include foreign fields into a text filter relevance (used for global search). Not sure whether your solution supports the relevance.

yurikuzn:
Here's where it's used.

Additionally. Removing public methods is not OK for non-major releases (for major releases is not advisable either). It's a quite major redesign.

mozkomor05:
Thank you for the feedback. I'll be busy for the next few (four?) days and I'll try to fix everything later.

mozkomor05:
I found some time. Global search tested and seems to work, few bugs fixed + unit and integration tests passed.

PR #2380, opened 2022-07-16
feature: grunt file watchers

superewald:
Add file watchers for grunt to increase build performance and development experience. 

## motivation

Currently a developer have to run `grunt build` to apply changes which triggers the whole build chain. This is very time consuming and leads to several unpleasant side effects.

## summary

- added grunt watchers using [grunt-contrib-watch](https://github.com/gruntjs/grunt-contrib-watch)
- changed `grunt dev` task which now triggers an initial build and start watch task afterwards

## usage

Simply run `grunt dev`

## details

The grunt watcher is configured to watch on changes of:

- theme styles (`frontend/**/*.less`)
- frontend js bundle (files specified in `frontend/bundle-config.json`)
- frontend libraries (`frontend/libs.json`)
- backend files (`['application/**', 'html/**', 'public/**', 'install/**', 'bin/**', '*.php', '.htaccess', 'web.config']`)
- backend composer libs (`composer.lock`)

The watchers itself will handle building/compiling and copying resulting files to build directory.

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -83,6 +83,8 @@ module.exports = grunt => {
         };
     });
 
+    grunt.lessData = lessData;
+
     grunt.initConfig({
         pkg: pkg,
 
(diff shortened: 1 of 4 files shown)


yurikuzn:
Do you use developer mode? I find a need to run `grunt less` only when changing less files.

superewald:
> Do you use developer mode? I find a need to run `grunt less` only when changing less files, it executes in a couple of seconds. You might be running Espo from the `build` directory?

Well, just realized I was going all the wrong way for development, somehow thought `grunt build` would be required. Sorry for that, I overread the Configuring for development section in docs..

In this case I'd adjust the watcher scripts to only trigger on files that need compilation/translation (`frontend/**/*.less`, `client/src/**/*.js`) and skip copying files to `build` directory.

PR #2383, opened 2022-07-18
Fixed max collection size

mozkomor05:
Prevented users from fetching full-size collection by setting `maxSize` as 0. (e.g. `/api/v1/Quote/action/list?maxSize=0`.)

Setting negative `maxSize` now also causes 403 error instead of 500.

### DIFF
--- a/application/Espo/Core/Record/SearchParamsFetcher.php
+++ b/application/Espo/Core/Record/SearchParamsFetcher.php
@@ -228,6 +228,10 @@ class SearchParamsFetcher
             $params['maxSize'] = $limit;
         }
 
+        if ($value < 1) {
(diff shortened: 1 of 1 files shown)


yurikuzn:
It's really should not treat 0 as 'select-max-allowed'. But it shouldn't throw an error either for a backward compatibility and consistency. E.g. one may make an API call to obtain a max-count and pass maxCount=0, they will get an error now.

yurikuzn:
Fix: https://github.com/espocrm/espocrm/commit/68ccfde3f30084bb394a2e1d8f66b08d23f0abf8

yurikuzn:
It turned out there were 2 bugs, in select manager and ORM delete and update queries. Thanks for reporting.

PR #2409, opened 2022-08-20
Shift select implementation

mozkomor05:
Allow user to select multiple records in `record/list` view by holding `Shift` key.

### DIFF
--- a/client/src/views/record/list.js
+++ b/client/src/views/record/list.js
@@ -512,13 +512,7 @@ function (Dep, MassActionHelper, ExportHelper) {
             'click .record-checkbox': function (e) {
                 const $target = $(e.currentTarget);
 
-                if (!this._lastChecked) {
-                    this._lastChecked = $target;
-
-                    return;
-                }
-
-                if (e.shiftKey) {
+                if (e.shiftKey && this._lastChecked) {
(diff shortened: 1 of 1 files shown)


yurikuzn:
Interesting idea. I will take a look.

v7.2 is frozen for feature adding, so I think we'd rather address it to v7.3.

Could also accept a CLA please. https://github.com/espocrm/cla

yurikuzn:
Maybe would be reasonable to use document.activeElement instead of storing last clicked element. Not sure yet, need to try first.

client/src/views/record/list.js
```diff
@@ -505,15 +510,31 @@ function (Dep, MassActionHelper, ExportHelper) {
              * @this module:views/record/list.Class
              */
             'click .record-checkbox': function (e) {
-                let $target = $(e.currentTarget);
+                const $target = $(e.currentTarget);
+
+                if (!this._lastChecked) {
```
yurikuzn, line 515: This if-block cause an issue when checked the first time that the item is not actually selected. Seems it's added by a mistake and should be removed. Then the following if to be changed to

```
if (e.shiftKey && this._lastChecked) {
```

mozkomor05: Good point, missed that.

PR #2412, opened 2022-08-23
Call `loadListAdditionalFields` on record service

mozkomor05:
Shouldn't `loadListAdditionalFields` method be called on the record service of the related entity?

This will not affect Field Processing, since List Load Processor loads classes based on `$entity->getEntityType()` . However in many cases developers might want to extend `loadListAdditionalFields`.

### DIFF
--- a/application/Espo/Core/Record/Service.php
+++ b/application/Espo/Core/Record/Service.php
@@ -1036,7 +1036,7 @@ class Service implements Crud,
             ->find();
 
         foreach ($collection as $itemEntity) {
-            $this->loadListAdditionalFields($itemEntity, $preparedSearchParams);
+            $recordService->loadListAdditionalFields($itemEntity, $preparedSearchParams);
 
             $recordService->prepareEntityForOutput($itemEntity);
         }


yurikuzn:
Extending by means of class inheritance is not recommended in Espo. That method should have been 'private' rather or copied to another private method.

I wander how PHPStan didn't catch that you calling a protected method in the PR. Might be a bug in PHPStan.

mozkomor05:
> Extending by means of class inheritance is not recommended in Espo. That method should have been 'private' rather or copied to another private method.
> 
> I wonder how PHPStan didn't catch that you calling a protected method in the PR. Might be a bug in PHPStan.

So do I, I feel stupid now. Somehow I even managed to run the code.

yurikuzn:
Maybe there're no relationship panels in your entity.

mozkomor05:
> Maybe there're no relationship panels in your entity.

There were. Turns out my record service redefined the `loadListAdditionalFields` method as public. Sorry for wasting your time anyway. I absolutely missed the fact that it is protected.

yurikuzn:
No problem. I recommend to rely on extending by means of our framework whenever  it's possible rather than extending. The ability to extend is still there for the backward compatibility sake only. It worked well when the app was small, when collisions made by multiple extensions where not a thing.

PR #2430, opened 2022-09-13
let json\retrieve works for stdclass

eymen-elkum:
This helped to play with data field on Note entity:

```
$fields = json\retrieve(data, 'fields');
record\update(parentType, parentId, 'description', array\join($fields, ' , '));
```


yurikuzn:
You can use it with `json\encode`.

Issue #2448, opened 2022-09-27
Enum's options declared in "actualFields" are object instead of array (enum declaration in custom field)


kczereczon:
https://github.com/espocrm/espocrm/blob/8e62ed2fb8e44d09ac8932a1afbd9230fa6c0c45/client/src/view-helper.js#L451

I've created custom field with enum declared in actualFields, when I tried to add enum to layout the following message shows up
`jQuery.Deferred exception: list.indexOf is not a function. (In 'list.indexOf(value)', 'list.indexOf' is undefined) (2)` 

![image](https://user-images.githubusercontent.com/22079016/192412183-a450fc3b-4b59-482c-a115-ec6d110ec011.png)
![image](https://user-images.githubusercontent.com/22079016/192412239-2348083e-b29e-4bd1-82d1-7c4182571634.png)

kczereczon:
Options are fetched as map even in defs

![image](https://user-images.githubusercontent.com/22079016/192413854-e30e5a3a-fac9-4704-b452-2baeaae64542.png)

![image](https://user-images.githubusercontent.com/22079016/192413575-b3f7c998-77a9-4a6e-86f1-f72711acfa97.png)

yurikuzn:
Could you debug deeper? I can't think of any place where the issue can happen.

yurikuzn:
I think I know where the issue happen. Will let you know. Maybe will need to fix in the master branch as the fix can be risky.

yurikuzn:
Fix: https://github.com/espocrm/espocrm/commit/cd92e4fcd86309c9ed3dec41cec5f3a88a4f69e2

PR #2449, opened 2022-09-27
#2448 map object to array when fetching options

kczereczon:
added conversion from object to array when argument is other than Array


yurikuzn:
It's a hack, issue should not be fixed this way.

PR #2453, opened 2022-09-28
Backbone model change fix

mozkomor05:
Model `set` method's first parameter shouldn't  be passed by reference, as it could cause a lot of very hard to debug and unexpected behaviour. E.g. because of how the changes are detected in the Backbone library, passing by reference causes Backbone not to detect attribute changes, thus `change` event is not called. 

Practical example of the bug:

- Create about 10 different teams with positions.
- In the link multiple field with roles at the user detail select all ten teams at once.
- Position column won't render.

Cause: Link multiple field is set by reference => change event won't get fired => `loadRoleList` won't get called.

### DIFF
--- a/client/src/views/fields/base.js
+++ b/client/src/views/fields/base.js
@@ -840,7 +840,7 @@ define('views/fields/base', ['view'], function (Dep) {
                 });
 
                 this.listenTo(this, 'change', () => {
-                    var attributes = this.fetch();
+                    const attributes = Espo.Utils.deepClone(this.fetch());
 
                     this.model.set(attributes, {ui: true});
                 });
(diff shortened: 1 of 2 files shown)


mozkomor05:
Another possible fix is to add the `cloneDeep` to the `fetch` function of link-multiple. However, I think that adding it right before the `set` is better as it prevents accidental bugs like this.

yurikuzn:
The `fetch` method is supposed to create newly create object w/o references inside. I'd better to fix the `fetch` implementation if problem fields. I feel not that comfortable to add excessive deep cloning.

yurikuzn:
Fix: https://github.com/espocrm/espocrm/commit/3e4c738ab1ec820d92c01ea5931af6f26a8c96fe

yurikuzn:
I applied the fix to the `fix` branch. Hope it won't bring any side effects.

PR #2460, opened 2022-10-02
Fixed model seed cache

mozkomor05:


yurikuzn:
This code never worked since the beginning, accepting in the fix release w/o a lot of testing is not an option. I was about to accept it.

mozkomor05:
> This code never worked since the beginning, accepting in the fix release w/o a lot of testing is not an option. I was about to accept it.

I don't get it. Is the fix wrong or should I PR into master?

yurikuzn:
It can really cause a lot of issues. I was so tired, so could accept it. luckely I didn't.

yurikuzn:
I see that it potentially could cause a lot of bugs.

mozkomor05:
Sure. I get it. Never commit untested (even if it was **_dead code_** previously) code to `fix`. So should I create PR to `master`?

PR #2461, opened 2022-10-02
added other reserved keywords

mozkomor05:
Recently a colleague of mine created an entity called "Object" and accidentaly killed the app.

The list of other reserved keywords: https://www.php.net/manual/en/reserved.other-reserved-words.php


yurikuzn:
Thanks. `enum` might be reserved too since PHP 8.1.

PR #2473, opened 2022-10-20
Fixed requirePromise in fix branch

mozkomor05:
Add `requirePromise` fix to *fix* branch, because the method is already published and it doesn't have much use, as it doesn't pass the loaded dependencies. 

The fix cannot even potentially break any existing Espo instance.

### DIFF
--- a/client/src/loader.js
+++ b/client/src/loader.js
@@ -766,7 +766,7 @@
             return new Promise((resolve, reject) => {
                 this.require(
                     subject,
-                    () => resolve(),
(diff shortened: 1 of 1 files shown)


yurikuzn:
I noticed that we have a wrong jsdoc parameter type. Should be like `{...string}`.

yurikuzn:
Note that promises does not support multiple arguments in resolve callback. We used to have a polyfill where it worked. I'll fix the function to support only one argument.

PR #2483, opened 2022-10-27
Stream: Update notes when edited (with web socket)

yurikuzn:


Kharg:
This is working now!
Thank you @yurikuzn 
One little tip, wouldn't it be better if it used {highlight: true} for edits as it does for the detail view? So users can notice if the note was edited.
Also what about fetching the stream if the note was deleted?

yurikuzn:
> wouldn't it be better if it used {highlight: true}
Did you test it. Would it work?

> Also what about fetching the stream if the note was deleted?
I don't have time for it.

PR #2490, opened 2022-11-01
Regex with optional capture groups might cause catastrophic backtracking

mozkomor05:
I noticed that when I uplaod large zip files, sometimes, the app returns error 500, instead of normal error output. 

I found out that the error is due to the [catasthropic backgracing](https://www.regular-expressions.info/catastrophic.html) which is not unusual with regex expressions containing lot of optional capture groups. The error causes `preg_replace` to return `null`, causing error since return value doesn't match.

### DIFF
--- a/application/Espo/Core/Api/ErrorOutput.php
+++ b/application/Espo/Core/Api/ErrorOutput.php
@@ -215,8 +215,7 @@ class ErrorOutput
 
     private function clearPasswords(string $string): string
     {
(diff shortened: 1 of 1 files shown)


yurikuzn:
Thanks. Does Espo write the whole input with the file contents to the log?

mozkomor05:
It doesn't. I had to wrap it around custom logs to catch the issue.

PR #2502, opened 2022-11-14
fix max depth

mozkomor05:


yurikuzn:
I need to revise. The fix will cause small issues.

PR #2514, opened 2022-11-19
css fix

mozkomor05:
When 25% < 80px, the currency field overflows. Since table-cells don't support min-width (min and max widths have undefined behaviour on tables since CSS 2.1), this can be resolved in one of two following ways:
- Remove the table layout and style the inputs using inline-block.
- Change table layout to `auto` and let the browser fix the overflowing width.

image

### DIFF
--- a/frontend/less/espo/custom.less
+++ b/frontend/less/espo/custom.less
@@ -3659,7 +3659,7 @@ a.close:hover {
 }
 
 .input-group-currency {
-    table-layout: fixed;
+    table-layout: auto;
     width: 100%;
 
     > .input-group-item:last-child {


yurikuzn:
Auto causes the issue:

image

yurikuzn:
Might need to remove
```
.selectize-input {
    min-width: 80px;
}
```

mozkomor05:
Does the `auto` issue apply to currency field?

yurikuzn:
Fixing it only for currency is not the way to go. The same issue is for all input-group fields having selectbox.

mozkomor05:
I also noticed that select boxes don't work inside `.table>tbody>tr>td` as the dropdown gets cutted by `overflow:hidden`.

mozkomor05:
> I also noticed that select boxes don't work inside `.list>table td` as the dropdown gets cutted by `overflow:hidden`.

That might actually be irrelevant. Maybe I should not put the fields inside the container with `list` class.

yurikuzn:
Bad. Need to find a solution. Selectbox already took me more than 30h of work.

yurikuzn:
Does it work in cases like on your screenshot in the first post?

mozkomor05:
The select box? No it doesn't, it cuts the content of the dropdown. But that might be my fault, since I render the field inside a table which is inside a container with the `list` class. Maybe I should modify my custom view to use different class and copy the styles I need.

yurikuzn:
I though it's the Quote items field from the Sales Pack.

For dropdowns on the right on the list we have overflow: visible as an exclusion. Also whoever customized list view to be editable will have the issue.

mozkomor05:
As far as I know, there's no way an edit-mode field can get inside a `record/list` view (with `list` class) in native Espo.

mozkomor05:
Yes. So the issue with `overflow:hidden` is my fault. 

There's still the issue when 25% of the width of the field is less than 80px.

yurikuzn:
I will remove 80px min width limitation that will fix the issue.

PR #2539, opened 2022-12-23
minimatch - security update

jakub-frajt:
Hi,

security update of minimatch library (https://github.com/advisories/GHSA-f8q6-p94x-37v3).


yurikuzn:
It's already fixed in the master. As it's not used in production, the only reason to have it fixes, is to suppress npm warnings.

PR #2548, opened 2023-01-02
Support more subquery operators

mozkomor05:
Support for multiple subquery operators (e.g. `>=s`, `<s`...).

I am aware that this will require a lot of testing, so take this more as a suggestion. 

Another suggestion: make is possible to select subquery. The current workaround is to create a function converter that strips the quotes around the expression:

```php
public function convert(string ...$argumentList): string
    {
        return "(" . trim($argumentList[0], "'") . ")";
    }
}
```


yurikuzn:
This implementation has a design problem that =s represents *IN sub-query*, but >s represents *> sub-query*. I will implement it in the future, plus it will needs tests and expression implementations.

yurikuzn:
https://github.com/espocrm/espocrm/commit/6edce56ca799bb56be7da06dc6b15c661892e595

PR #2549, opened 2023-01-02
Added support for global mass actions defs

mozkomor05:
The list of mass actions is obtained from the global clientDefs, but the definitions are only obtained from the scope clientDefs, making it impossible to define global mass action with custom handler.


yurikuzn:
Thanks. This was missed when implemented. I'll add `massActionDefs` protected property to the class, as we need all props be defined.

Issue #2571, opened 2023-01-19
Call dropdown UI issue when permission applied


rabiibrahimi:
**Describe the bug**
When a user has the permission to view his team's calls and only edit his own. when accessing a call of his team the dropdown is still shown as empty. 

**To Reproduce**
1. create a regular user (set permission to create call | view his team | edit his own | delete No)
2. assign a call to his team 
3. login in as the regular user access the call assigned to his team on top left dropdown is shown with empty element. 
...

**Expected behavior**
i expect the user to only see the call and dropdown should't appear. 

**Screenshots**

**EspoCRM version**
version 7.3.2
<img width="966" alt="Screenshot 2023-01-19 at 10 39 04" src="https://user-images.githubusercontent.com/19490674/213421158-8318d00e-0459-42e7-baf3-db3b5c6a2970.png">

**Additional context**

yurikuzn:
I found the mistake: https://github.com/espocrm/espocrm/commit/72a1d73848a409d76352d6a68e62c7c6010090ad

rabiibrahimi:
great 
thanks

PR #2572, opened 2023-01-19
Add checkField ACL for the call

rabiibrahimi:
Added ACL  checkField on call detail view to present dropdownActions to appear is user doesn't have permission to edit.

fix for https://github.com/espocrm/espocrm/issues/2571


yurikuzn:
On that stage it's not possible to know ACL as some record data yet to be loaded.

The issue is not with ACL but with the logic that hides empty dropdown.

yurikuzn:
This https://github.com/espocrm/espocrm/blob/7.3.2/client/src/views/record/detail.js#L874 might be not called after actions are hidden. Maybe not yet rendered.

PR #2574, opened 2023-01-20
fix convert lead map fields

eymen-elkum:
When tried to use the convert fields map feature I faced this issue and fixed it.
Not sure if I have to pull to master or fix branches, you can edit the PR to change the target branch

photo_2023-01-20 3 07 57 PM

### DIFF
--- a/application/Espo/Modules/Crm/Tools/Lead/ConvertService.php
+++ b/application/Espo/Modules/Crm/Tools/Lead/ConvertService.php
@@ -232,7 +232,7 @@ class ConvertService
             ->find();
 
         foreach ($meetings as $meeting) {
-            if ($contact &&  $contact->hasId()) {
+            if ($contact && $contact->hasId()) {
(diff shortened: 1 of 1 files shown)


yurikuzn:
More detail input about the issue with PRs is needed.

eymen-elkum:
In the [line](https://github.com/espocrm/espocrm/pull/2574/files#diff-f434b574881b17f581334c17a2df2149eb71bebe28c30d82838342db5c6a578cR404) we need type of field on the target scope (ex: RealEstateproperty) but it search in Lead scope, so the type  is empty always unless the same field is exist on both entities.

yurikuzn:
It's clear now. Thanks for explanation.

yurikuzn:
On your screenshot, there are currency fields. They were not affected by the bug, right?

eymen-elkum:
Not affected

Issue #2546, opened 2022-12-31
Database helper shouldn't contain MySQL-specific sql


SuchAFuriousDeath:
**Is your feature request related to a problem? Please describe.**
A bit of a follow-up on #2537 ( I'm trying to make EspoCRM work with Postgres). In the class Espo\Core\Utils\Database, calling the methods 'doesSupportFulltext' and 'getMaxIndexLength' causes the SQL in the 'getTableEngine' method to be run, causing an error (Postgres has no such thing as engines, so the SQL fails). 

**Describe the solution you'd like**
Factor the platform-specific SQL into another class, perhaps make the current helper abstract and make child classes for each platform.

**Describe alternatives you've considered**
The Helper can be probably replaced the same way you suggested in the previous issue (#2537). I just don't consider it very nice and thought you might want to take a look at this, since you fullfilled my previous feature request.

SuchAFuriousDeath:
Also, if one wanted to add a Platform and Driver for Postgres, the helper would have to be overridden in the same inconvenient way.

yurikuzn:
I made some changes, supposed to work with `database.platform => 'Potsgresql'` now. But I didn't test. Hardcoded w/o introducing a new interface.

SuchAFuriousDeath:
@yurikuzn Now it throws a new error. I believe it's these 2 lines: 

https://github.com/espocrm/espocrm/blob/ac667356c528d0764e78dc6406dd09498deb6a1d/application/Espo/Core/Utils/Database/Helper.php#L140-L142

Reading the Driver code, it looks like it wants the version in the format returned by the method 'getDatabaseVersion' (for example 13.8.0) as opposed to the full version.

yurikuzn:
Not an easy fix. As it's opposite for MySQL and MardiaDB.

SuchAFuriousDeath:
Isn't it just a condition?

yurikuzn:
Please try `getDatabaseVersion` and let me know. I will be able to look not today.

yurikuzn:
It's too much of hardcode. Eventually, it should be refactory.

SuchAFuriousDeath:
@yurikuzn 
Right, I understand. Before your commit regarding the first issue, I already got further with the module. I got to the point where more than 100 tables were created (Now I'm writing the module from scratch based on the new 7.4 commit). From what I can remember, these are the issues I came across:

[...]

PR #2577, opened 2023-01-21
Postgres.

SuchAFuriousDeath:
Related to: #2546 
I got it to a point where you can log in and create, read and update some entities. Activities don't work and a lot of other things probably don't as well, just doing this to provide an overview of the issues so they can be solved better. I put comments in the code and started them with 'HERE', so look for that. Some changes are in the Postgres module because that's where I initially put them. Thank you for the time you have given my issues.


yurikuzn:
Great. I will take a look when get a chance. Currently I don't have time, got electricity only for a couple of hours per day.

application/Espo/Modules/Postgres/Core/Utils/Database/Orm/Fields/PersonName.php
```diff
@@ -0,0 +1,25 @@
...
+
+            if (empty($rowItem)) {
+                continue;
+            }
+
+            if ($alias) {
+                $item = $alias . '.' . $item;
+            }
+
+            $item = "COALESCE:({$item}, '')"; // HERE - change IFNULL to COALESCE, it can probably be done in a better way, not sure
```
yurikuzn, line 20: As NULLIF is a part of Espo and is documented changing it here is not enough. It's needed to be converted on ORM level.

application/Espo/Core/Utils/Database/Schema/Converter.php
```diff
@@ -102,13 +102,14 @@ class Converter
...
-        Metadata $metadata,
-        FileManager $fileManager,
-        Schema $databaseSchema,
-        Config $config,
-        Log $log,
+        Metadata     $metadata,
+        FileManager  $fileManager,
+        Schema       $databaseSchema,
+        Config       $config,
+        Log          $log,
```
yurikuzn, line 109: Would be good if automatic argument aligning and other code style fixes wouldn't have been enabled in IDE. So only actual changes were in PRs.

application/Espo/Modules/Postgres/Core/Utils/Database/Orm/Fields/PersonName.php
SuchAFuriousDeath: I thought so. But it's not NULLIF, it's IFNULL.. Pg actually has NULLIF.

application/Espo/Core/Utils/Database/Schema/Converter.php
SuchAFuriousDeath: Sorry about that.

application/Espo/Modules/Postgres/ORM/QueryComposer/PostgresqlQueryComposer.php
```diff
@@ -0,0 +1,808 @@
...
+            }
+
+            if (strpos($attribute, '.') > 0) {
+                [$alias, $attribute] = explode('.', $attribute);
+
+                $column = $this->toDb($this->sanitize($attribute));
+                $left = $column;
+            } else {
+                $column = $this->toDb($this->sanitize($attribute));
+                $left = $column;
```
yurikuzn, line 271: You missed alias/table on purpose here?

application/Espo/Modules/Postgres/ORM/QueryComposer/PostgresqlQueryComposer.php
```diff
@@ -0,0 +1,808 @@
...
+                        $params,
+                        $noCustom
+                    );
+                }
+            }
+
+            return implode(", ", $arr);
+        }
+
+        if (str_starts_with($orderBy, 'LIST:')) { // HERE - same thing as above, no field function
```
yurikuzn, line 220: I refactored this so that `POSITION_IN_LIST` complex expression function is used instead. `LIST:` is a legacy.

https://github.com/espocrm/espocrm/commit/d82c297b797b5a38d98749f202044403ed1b74f6

application/Espo/Modules/Postgres/ORM/QueryComposer/PostgresqlQueryComposer.php
```diff
@@ -0,0 +1,808 @@
...
+
+        if (Util::isComplexExpression($field)) {
+            $leftPart = $this->convertComplexExpression($entity, $field, false, $params);
+
+            $isComplex = true;
+        }
+
+        if (!$isComplex) {
+            if (!$entity->hasAttribute($field)) {
+                return '0'; // HERE - probable rewrite to $this->quote(false)
```
yurikuzn, line 593: Done for all '0', '1'.

application/Espo/Modules/Postgres/ORM/QueryComposer/PostgresqlQueryComposer.php
```diff
@@ -0,0 +1,808 @@
...
+                            );
+                        } else {
+                            $leftPart = $this->getAttributePath($entity, $field, $params);
+                        }
+                    }
+                }
+            } else {
+                $fromAlias = $this->getFromAlias($params, $entity->getEntityType());
+
+                $leftPart = $this->quoteColumn($fromAlias . '.' . $this->toDb($this->sanitize($field))); //HERE - quoting needed
```
yurikuzn, line 727: Fixed.

application/Espo/Modules/Postgres/ORM/QueryComposer/PostgresqlQueryComposer.php
```diff
@@ -0,0 +1,808 @@
...
+
+        if ($alias) {
+            $sql .= " AS " . $this->quoteIdentifier($alias);
+        }
+
+        /*if (!empty($indexKeyList)) { //HERE - index hinting not supported
+            foreach ($indexKeyList as $index) {
+                $sql .= " USE INDEX (" . $this->quoteIdentifier($this->sanitizeIndexName($index)) . ")";
+            }
+        }*/
```
yurikuzn, line 123: Added `indexHints` property to be able to disable in an inherited class.

SuchAFuriousDeath:
@yurikuzn I've been watching the commits in your personal fork and they look very good. Thank you so much, it must've taken a lot of work and mental effort. If I may ask, have you already thought about the non-existing functions like IFNULL and FIELD (I actually found even more recently). Will you add an abstraction layer, or should it just be done with overriding the 'getFunctionPart' in the QueryComposer?

SuchAFuriousDeath:
@yurikuzn What about the DBAL FieldTypes, though? The LongText and Text (and JsonArray and JsonObject types that extend from it) types have a hardcoded SQL declaration to LONGTEXT and TEXT respectively, non of which seem to exist in PG. only the TEXT type exists.

SuchAFuriousDeath:
And by the way, you said something about converting and empty string to null in expressions... but expressions aren't used in the place I was talking about. So, maybe do something like Expression::emptyDate() ? Not sure how I would do it.

https://github.com/espocrm/espocrm/blob/1ce38253382eeb483a9da77cae1c11b3f2121d81/application/Espo/Modules/Crm/Tools/Activities/Service.php#L1094-L1109

yurikuzn:
Did you try to replace them with a string `'null'`?

SuchAFuriousDeath:
I'll try and let you know.

SuchAFuriousDeath:
@yurikuzn It worked.

SuchAFuriousDeath:
The same issue was also in the Calendar service class and seems to be fixed in the same way. But I found another thing. The ON DUPLICATE clause in insert seems to work a bit differently in pg. First of all, it's 'ON CONFLICT' and second of all, it seems to me that if you want to update (alternatively you can 'DO NOTHING'), you have to specify the columns, on which the conflict happens or something like that. I'll try to do something about it myself.

SuchAFuriousDeath:
But the SQL declarations are a bigger problem.

yurikuzn:
Maybe to add `id` and all columns from `unique` indexes of the current entity.

SuchAFuriousDeath:
I'll give it a shot.

yurikuzn:
I added a commit that adds indexes to ORM metadata for relationship entities. These indexes will be needed to be able to get columns for INSERT-UPDATE when dealing with relationship tables.

SuchAFuriousDeath:
@yurikuzn Amazing. But what about the MEDIUMTEXT issue?

yurikuzn:
You need to choose a corresponding type and length in the column preparator implementation.

yurikuzn:
Or this causing the problem? https://github.com/espocrm/espocrm/blob/master/application/Espo/Core/Utils/Database/DBAL/FieldTypes/TextType.php

SuchAFuriousDeath:
Yes, that. It's hardcoded to MEDIUMTEXT, but PG only has TEXT.

SuchAFuriousDeath:
JsonArray and jsonObject are fine, they can be mapped to the default json dbal type, but text can't.

yurikuzn:
I'll fix it.

SuchAFuriousDeath:
Tried with the new commit and it works.

SuchAFuriousDeath:
@yurikuzn actually, found one more thing.

https://github.com/espocrm/espocrm/blob/282f701b7761aeffbdde4d0168c0744a5d3a0e4d/application/Espo/Repositories/EmailAddress.php#L334

SuchAFuriousDeath:
@yurikuzn By the way the 'ON CONFLICT' clause can be solved like this, I think.
`
    protected function composeInsertQuery(
        string  $table,
        string  $columns,
        string  $values,
        ?string $update = null
    ): string
    {
        $tableQuoted = $this->quoteIdentifier($table);

        $sql = "INSERT INTO $tableQuoted ({$columns}) {$values}";

        $constraint = $this->quoteIdentifier("{$table}_pkey");

        if ($update) {
            $sql .= " ON CONFLICT ON CONSTRAINT $constraint DO UPDATE SET $update";
        }

        return $sql;
    }
`
I am not aware of any other issues at the moment, except fulltext. If I may ask, all this refactoring you've made - is it coming in 7.4?

yurikuzn:
Unique key is necessary, otherwise it will break the app. Refactoring is already in master.

SuchAFuriousDeath:
You're probably right. I'll have to grab the first unique index from metadata. The primary key implies the use of a unique index. but it's the wrong one in N-N relation middle tables.

yurikuzn:
A fulltext support may be a no-brainer, just convering MATCH complex experssion in the query composer.

SuchAFuriousDeath:
But to make it fast, Probably some kind of special indexes will have to be used, no?

SuchAFuriousDeath:
By the way i tried importing my emails and ran into index size issues. Not really sure what to do about it.

SuchAFuriousDeath:
Could you maybe move some index logic to the index helper? Maybe we could create a separate fulltext index for each of the columns, instead of one for all.

SuchAFuriousDeath:
I had no idea that syncing would close the issue, that's funny.

yurikuzn:
Maybe create a method in the IndexHelper that will create index definition(s) in ORM metadata for full-text specificaly.

SuchAFuriousDeath:
I don't think I understand DBAL well enough at the moment. I tried the very platform hack you link to, but there's and issue... When the index doesn't exist, it's created properly, but when it does exist, it tries to create it again (Which of course throws an error). I think it's something about the index not being defined on plain columns, but rather on the results of a function called on a column. It messes DBAl up somewhere, because it probably relies on the indexes just being on plain columns. And I don't really know how to solve it, because I only have a very basic understanding of DBAL... maybe it's possible to somehow prevent DBAL from trying to create the index again. The index you suggest will probably work for Espo, that I agree with.

yurikuzn:
It could be that DBAL didn't recognize that existing index is the same and tries to create a new one. Maybe because it differs. See in the Comparator class in DBAL.

SuchAFuriousDeath:
DBAL doesn't put the index in the schema for some reason.

SuchAFuriousDeath:
It's not the comparator, the index simply isn't is the schema :/

SuchAFuriousDeath:
This part is the problematic part. 
image
image
$indexColumns will be empty and nothing gets passed forward. The thing is, even if we somehow forced the columns in by creating a custom Schema Manager and overriding the function, who knows what consequences it would have... The only feasible solution that comes to mind is to simply figure out a way to make DBAL ignore fulltext indexes and then maybe drop and re-create them all during a hard rebuild. By the way, maybe you could add something like 'hardRebuildActionList' in the metadata (And process those during a hard rebuild). Could be useful, not only for this issue.

yurikuzn:
I'll think about it.

I create a PR https://github.com/espocrm/espocrm/pull/2599. Not sure whether GitHub would allow me to add you as a collaborator of the PR.

Or maybe it would be better to create a separate branch.

yurikuzn:
I wanted to check how DBAL works with Postgres, but Espo couldn't run rebuild w/o ORM working, so I had to write an implementation.

What I discovered, is that SEQUENCEs (auto-increment) being dropped and created (every second run). This issue has to be resolved as well. I will investigate it more.

SuchAFuriousDeath:
That actually happened to me too.

SuchAFuriousDeath:
@yurikuzn Here is what i have so far (But in a module). Didn't know the issue can be reopened with new commits, I'm not that experienced with git, so sorry for the mess.

SuchAFuriousDeath:
- composeUpdateQuery: probably utilize CTEs or subquery to handle limit and orderBy.
- composeInsertQuery: I'm not sure how to get the columns for the ON CONFLICT part.

yurikuzn:
Dropped sequence issue fix: https://github.com/espocrm/espocrm/commit/4598b58fbb73df7bf28413cd82abd420442ed136

yurikuzn:
Insert on conflict: https://github.com/espocrm/espocrm/pull/2599/commits/2ec385f4b58627921e417b44e324a0b1b87fee16

SuchAFuriousDeath:
I completely forgot that we can modify the diff. That could probably be used to solve the fulltext indexes.

yurikuzn:
I did some performance testing for Espo with Postgres. Searching with linked-with filter is pretty fast (as so in MariaDB, but very slow in MySQL) as `WHERE id IN (subquery)` is optimized well. But searching with not-linked-filter is extremely slow (it's not that slow in both MariaDB and MySQL), I had to end the process manually as my PC was about to catch on fire. Turned out that PostgreSQL has no optimization for  `WHERE id NOT IN (subquery)`.  What is weird that it seems there was no any timeout triggered. Might need additional configuration. Otherwise, it would be easy to DoS the server.

I think I will have to implement `EXIST`/`NOT EXISTS` operators in ORM.

SuchAFuriousDeath:
Oh. Interesting.

yurikuzn:
I implemented full-text index reverse engineering with a custom Schema Manager. Also implemented full-text search, but not yet sure whether it works appropriately. Didn't test on big tables with real data.

SuchAFuriousDeath:
Wow. Great work.

SuchAFuriousDeath:
@yurikuzn You're still missing an IF implementation, I think it should look roughly like this (maybe with a check for argument count)

```php
        if ($function === 'IF'){
            $condition = $argumentPartList[0];
            $then = $argumentPartList[1];
            $else = $argumentPartList[2];

            return "CASE WHEN {$condition} THEN {$then} ELSE {$else} END";
        }
```

SuchAFuriousDeath:
@yurikuzn Are the limit and order clauses actually utilized in an update statement somewhere in Espo?

yurikuzn:
It would be problematic to implement UPDATE where joined table column is present the SET expression. Maybe to leave it unimplemented as it's not used in Espo.

PR #2603, opened 2023-02-13
Update select-records.js

rabiibrahimi:
Add ability to show pagination one select records modal (if pagination enabled).

### DIFF
--- a/client/src/views/modals/select-records.js
+++ b/client/src/views/modals/select-records.js
@@ -255,6 +255,9 @@ define('views/modals/select-records', ['views/modal', 'search-manager'], functio
                 checkAllResultDisabled: !this.massRelateEnabled,
                 buttonsDisabled: true,
                 skipBuildRows: true,
+                pagination: this.getConfig().get('listPagination') ||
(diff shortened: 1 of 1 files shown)


client/src/views/modals/select-records.js
```diff
@@ -255,6 +255,9 @@ define('views/modals/select-records', ['views/modal', 'search-manager'], functio
                 checkAllResultDisabled: !this.massRelateEnabled,
                 buttonsDisabled: true,
                 skipBuildRows: true,
+                pagination: this.getConfig().get('listPagination') ||
+                    this.getMetadata().get(['clientDefs', this.scope, 'listPagination']) ||
```
yurikuzn, line 259: I'd rather have it w/o clientDefs. So that clientDefs > listPagination was applicable only for the list view.

rabiibrahimi: I didn't pa attention to that 
sorry

yurikuzn: No problem. I merged.

rabiibrahimi: Thanks

PR #2610, opened 2023-02-16
fix foreign enum & array with optionPath & translations

eymen-elkum:
Hi Yuri,

This is an issue when you create a foreign field for enum, array or multiple enum fields with optionsPath or translation set on field defs directly as path.
I hope it is useful


yurikuzn:
Hi,

Fix is fine. The only thing needed, is that we need to clone translations with `Espo.Utils.clone(...)`  before assigning it to something.

eymen-elkum:
I left your part as it is, so you use clone if found it needed

yurikuzn:
I meant cloning translations Object. I will do after merge.

yurikuzn:
I confused translation with translatedOptions. But you did right that you cloned options.

PR #2629, opened 2023-03-02
Appliers not applied for kanban requests

eymen-elkum:


yurikuzn:
This functionality was not the best decision and was added in an urgency. I wouldn't like it propagating to the application from out of Record/Service. Maybe we will need to add the similar functionality defined in metadata > selectDefs and applied automatically. Not yet sure.

yurikuzn:
JFI. Usage of InjectableFactory is not needed here, as the class itself can be injected via constructor. If's used in Record/Service only because of constraints dictated by a legacy.

eymen-elkum:
Thank you for the clarification, please let me know the alternative way when possible because we are using this in production project as additional access layer.

yurikuzn:
Maybe this will work for you: https://forum.espocrm.com/forum/developer-help/88614-apply-default-filter-using-accesscontrolfilter?p=88643#post88643

eymen-elkum:
Sorry for the late reply, 

Access Control Filters are applied per scope, is there a way to make them applied globally?
For many projects we needed to have additional generic filters and making it possible to have a class filter to "Global" will be perfect!

yurikuzn:
There's no way besides binding a custom FilterFactory here: https://github.com/espocrm/espocrm/blob/7.3.3/application/Espo/Core/Select/AccessControl/Applier.php#L62. But this would not be 100% future proof way as an implementation may change.

Adding new customization abilities to the framework time investments (even if it's a few lines of code). Currently, I'm concetrated on another work, don't want to make bad design decisions.

PR #2632, opened 2023-03-04
bug fix: tcpdf total page numbers for groups

bandtank:
The total page number was not correct for groups. The page number for a non-group was used, which produced the wrong result. The replaced value for the `totalPageNumber` placeholder in group mode has been changed from `ptp` to `ptg`. The `setPageNumbers` function has also been introduced to enhance code reuse.

Here is how the value of `totalPageNumber` changes as a result of the pull request:
```
Group   Page    Old   New
-----   ----    ---   ---
1       1       6     3
1       2       6     3
1       3       6     3
2       1       6     2
2       2       6     2
3       1       6     1
```
See [here](https://github.com/tecnickcom/TCPDF/blob/e3cffc9bcbc76e89e167e9eb0bbda0cab7518459/include/tcpdf_static.php#L72) for the TCPDF constant.

### DIFF
--- a/application/Espo/Tools/Pdf/Tcpdf/Tcpdf.php
+++ b/application/Espo/Tools/Pdf/Tcpdf/Tcpdf.php
@@ -128,20 +128,7 @@ class Tcpdf extends TcpdfOriginal
 
         $html = $this->headerHtml;
 
-        if ($this->useGroupNumbers) {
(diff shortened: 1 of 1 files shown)


yurikuzn:
Thanks. Could you access [CLA](https://github.com/espocrm/espocrm#contributing).

bandtank:
Done

PR #2642, opened 2023-03-06
Fix formatter enum translation

eymen-elkum:
Hi @yurikuzn 
Another issue with the same topic (enum with translation path), I also improved the last code I made on last pr


yurikuzn:
Can't take a look today as very busy. Please could you not to apply code style changes in PRs.

Issue #2644, opened 2023-03-06
Double consecutive `createView` calls may cause side effects


mozkomor05:
**Describe the bug**
Double consecutive `createView` calls may cause side effects (dangling promises) because of the way Bullbone works. Most often such an error is related to double consecutive re-render (closely related to https://github.com/espocrm/espocrm/issues/2635) or `prepare` call.

**Explanation**
Suppose (i) a view _A_ that creates nested view _B_ inside its `prepare` function. For example A = `time-ranges` and B = `item-detail` of working calendar field, which already exist in EspoCRM. Suppose (ii) that view B takes some time to load. We can simulate this by adding 
```js
this.wait(new Promise(resolve => setTimeout(resolve, 500)));
```
to its setup function. (I couldn't find a view in Espo that meets both conditions (i) and (ii), but developers can create such views. For sake of this example I appended the `wait` line to the `item-detail` manually).

Now:
1. Begin inline editation of view _A_
2. Make changes
3. **Close** inline editation

https://github.com/espocrm/espocrm/blob/6afd616e42deb6ec9f9590803fd1825fc28c01a5/client/src/views/fields/base.js#L1280-L1286

4. This causes a call to `setDetailMode` function, which calls `prepare`. This creates a promise _X_ on which depends the correct termination of inline editation.
5. Initial attributes are stored back to model, which causes change event. Causing another call to `prepare`.
6. Another call to `prepare` marks promise _X_ as cancelled. However, as view _B_ takes some time to load, promise _X_ haven't resolved yet.
7. This creates dangling promise _X_, thus inline editation never terminates correctly and editation links never disappear.

I think that this is fundamental issue with Bullbone, as developes have no feedback when createView terminates. I can think of adding a new option to `createView`, that tells `createView` to reject, when promise is marked as cancelled.

```diff
delete this._viewPromiseHash[key];

if (promise && promise._isToCancel) {
+    if (options.rejectOnCancel) {
+       reject();
+    }
    return;
}
```

yurikuzn:
The issue with closing inline-edit should be addressed anyway.

I believe changing the method to the following might work:

```js
            let promise = this.setDetailMode()
                .then(() => {
                    if (!noReset) {
                        this.model.set(this.initialAttributes, {skipRerender: true});
                    }

                    return Promise.resolve();
                })
                .then(() => this.reRender(true))
                .then(() => this.removeInlineEditLinks());
```

yurikuzn:
It simpler:

[...]

PR #2653, opened 2023-03-08
Fix inline edit close

mozkomor05:
Changing one field may cause other changes (e.g. via dynamic handler or dynamic logic) that are not reverted back to the UI when closing inline edit due to `skipReRender`. 

NOTE: This does not revert https://github.com/espocrm/espocrm/issues/2644 as `model.set` is executed before `setDetailMode` which is enough to avoid side effects described in the issue.


yurikuzn:
Thanks. Maybe I'll add skipReRenderInEditMode for this.

PR #2662, opened 2023-03-15
Order Extension by name

SuchAFuriousDeath:
Having extensions ordered by name makes it much faster to search in if you want to know whether a specific extension is installed.

### DIFF
--- a/application/Espo/Resources/metadata/entityDefs/Extension.json
+++ b/application/Espo/Resources/metadata/entityDefs/Extension.json
@@ -53,7 +53,7 @@
         }
     },
     "collection": {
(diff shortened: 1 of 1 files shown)


yurikuzn:
It orders by lastly updated extension that I find useful.

SuchAFuriousDeath:
I don't find it that useful, but I respect your opinion. What about at least making it sortable? Sorting is disabled in the layout by default.

yurikuzn:
I'll do 'name' sortable. Hope it will work.

SuchAFuriousDeath:
Great. Thank you for the quick response.

PR #2663, opened 2023-03-15
Fix removing row from record/list

mozkomor05:
`removeRowHtml` might re-render the view, so it must come at the end of the function. Otherwise, it will try to re-render the last deleted row. This might cause error in console if the view manipulates DOM in after render.

**Step-by-step reproduction:**
1. Create a custom field that requires valid element in `afterRender`. Example:
```js
define('custom:views/test-abc/fields/name', ['views/fields/varchar'], function (Dep) {
    return Dep.extend({
        afterRender: function () {
            Dep.prototype.afterRender.call(this);

            console.log(this.$el.length); // expecting 1
        }
    });
});
```
2. Go to a list view with single record.
3. Remove the record.
image

NOTE: An example of a view that throws an error if an element does not exist is e.g. edit mode of fields/int, where the AutoNumeric library requires a valid element.

### DIFF
--- a/client/src/views/record/list.js
+++ b/client/src/views/record/list.js
@@ -3181,8 +3181,6 @@ function (Dep, MassActionHelper, ExportHelper, RecordModal) {
                 this.checkedList.splice(index, 1);
             }
 
-            this.removeRowHtml(id);
-
             let key = id;
 
             this.clearView(key);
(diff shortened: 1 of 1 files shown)


yurikuzn:
Hi David. I wrote you an email yesterday, regarding CLA update. Not sure whether you received it. I asked about accepting our new CLA. Example: https://github.com/espocrm/cla/pull/78/files

PR #2667, opened 2023-03-18
Fixed duplicate key in translation

SuchAFuriousDeath:
Line 112 already contains the key with the same value.


yurikuzn:
Could you accept CLA please. https://github.com/espocrm/cla

yurikuzn:
Language files (except en_US) are generated when we import languages from POEditor. JFI. Not sure how duplicate appeared here, might be it was in the original when the file was generated last time.

PR #2673, opened 2023-03-24
add proxy support in Upgrade.php

hi-ko:
I didn't find a way to use a http proxy for command upgrade so I added proxy support in application/Espo/Core/Console/Commands/Upgrade.php

### DIFF
--- a/application/Espo/Core/Console/Commands/Upgrade.php
+++ b/application/Espo/Core/Console/Commands/Upgrade.php
@@ -439,6 +439,13 @@ class Upgrade implements Command
         curl_setopt($ch, \CURLOPT_RETURNTRANSFER, true);
         curl_setopt($ch, \CURLOPT_URL, $url);
 
+        // proxy support
+        $proxy = null;
+        $proxy = $this->config->get('proxyUrl', $proxy);
+        $proxyauth = $this->config->get('proxyAuth', $proxyauth);
+        curl_setopt($ch, CURLOPT_PROXY, $proxy);
+        curl_setopt($ch, CURLOPT_PROXYUSERPWD, $proxyauth);
+
         $result = curl_exec($ch);
 
         curl_close($ch);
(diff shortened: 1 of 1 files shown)


yurikuzn:
Such changes are extremely risky would require a lot of testing. In the future we need to come up some generic solution.

yurikuzn:
JFI, you can also upgrade manually, but downloading an upgrade package from the webstite.

hi-ko:
Why is adding proxy support in downloadFile / getVersionInfo per se risky? I would assume the other way around: upgrade should not be run without a proxy from a server ;-)
Maybe the whole Upgrade should not be owned and run by the runtime user (e.g. www-data) but that's another story ...

yurikuzn:
Any change in upgrade is risky that requires our QA to engage. If upgrade fails for some users, we have a huge headache. 

Adding new config parameters is the process that can't reverted. We also add all new parameters to docs. If we add these parameters for upgrade we would also need to support them in any CURL usage. We need some generic framework. We need to choose good names for such parameter either.

yurikuzn:
> upgrade should not be run without a proxy from a server

What do you mean?

hi-ko:
A web server should not have direct access to the internet. Best practice for security is to use a http proxy and monitor / whitelist urls. Similar how this is done with apt on debian/ubuntu.

> Any change in upgrade is risky that requires our QA to engage. If upgrade fails for some users, we have a huge headache.
So you mean I should create a feature issue instead, so your team can cover all the required steps?

yurikuzn:
We need a generic solution for all CURL usages. I added it to our backlog.

hi-ko:
good point! I already tried that but may have lost the env variables using sudo.

The way it works without the need to add the parameters to `/etc/profile`:
```
export http_proxy='http://1.2.3.4:8888'
export https_proxy=$http_proxy
sudo -E -u www-data php /var/www/html/espocrm/command.php upgrade
```

hi-ko:
or even a bullet proof fix would be to create a simple wrapper to be owned and run as root:

```
# cat ~/bin/esporcrm-command
export http_proxy='USERNAME:PASSWORD@PROXYIP:PROXYPORT'
export https_proxy=$http_proxy
export no_proxy='*.mycompany.com *.mycompany.local'
sudo -E -u www-data php /var/www/html/espocrm/command.php $@
```
```
# ~/bin/esporcrm-command upgrade
There are no available upgrades.
```

yurikuzn:
BTW upgrade is OK to run under root user.

yurikuzn:
When running under root, upgrade script applies *chown* for necessary files.

PR #2676, opened 2023-03-27
border-radios fixes for rtl theme

eymen-elkum:
This is for help with enabling border radios for rtl themes, current behaviour is zero border radios whereas all other themes has it set to 5px witch looks better.

Before:
Screenshot 2023-03-27 at 1 59 39 PM
After:
Screenshot 2023-03-27 at 1 58 50 PM

### DIFF
--- a/frontend/less/espo-rtl/custom.less
+++ b/frontend/less/espo-rtl/custom.less
@@ -356,17 +356,82 @@ table.table-admin-panel tr > td:first-child > div > a {
     padding-left: 10px;
 }
 
+.input-group > .input-group-btn:not(:first-child) > .btn:first-child {
+    border-top-right-radius: 0 !important;
+    border-bottom-right-radius: 0 !important;
+    border-top-left-radius: var(--border-radius) !important;
+    border-bottom-left-radius: var(--border-radius) !important;
+}
+
+.input-group > .input-group-btn:last-child > .form-control:last-child {
+    border-top-left-radius: var(--border-radius) !important;
+    border-bottom-left-radius: var(--border-radius) !important;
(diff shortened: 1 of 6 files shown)


eymen-elkum:
I strongly suggest that processing rtl made in different way, similar to how side/top navbar is processed.

The less files for RTL already can work as additional css just like a theme, so we can make this file generated separately and add it to the html when this option "RTL" is enabled.

Will be even more powerful if we added in defs the RTL languages and made crm automatically add the espo-rtl.css conditionally when user language is one of them.

eymen-elkum:
> Are changes in non-RTL less files necessary? I think it could be done w/o changing them.

Yes for global search round borders

yurikuzn:
There must be yet many other border radius issues other than these.

eymen-elkum:
I made some tests and not found till now, may need more tests, but this PR is for sure needed step on the way to fix all rounded borders

yurikuzn:
Could be with quick detail view modals. What came to my mind first.

eymen-elkum:
Just found other issues, will continue fixing and push here

frontend/less/espo-rtl/custom.less
```diff
@@ -356,15 +356,82 @@ table.table-admin-panel tr > td:first-child > div > a {
     padding-left: 10px;
 }
 
+.input-group > .input-group-btn:not(:first-child) > .btn:first-child {
```
eymen-elkum, line 359: ref: https://github.com/espocrm/espocrm/blob/master/frontend/less/espo/elements/form.less#L51

yurikuzn:
Looks good. Thanks.

PR #2678, opened 2023-03-28
RTL vertical theme

eymen-elkum:
I made this from 2 commit to help you follow what is the difference with the original less file, this helped me too, I skipped (not included) the styles of documents (tree list view) because I was not sure about them.


eymen-elkum:
squashed into one commit and made another branch for reference of separated commits:
https://github.com/eymen-elkum/espocrm/tree/rtl-vertical-separated-commits

PR #2702, opened 2023-04-19
fix prepare-lib.js

eymen-elkum:
I had an error when called this command, that dest file is not exist, bootstrap-datepicker for example


yurikuzn:
I'm not sure. We use this in daily basis.

yurikuzn:
Likely, it's not supposed to be run separately. I'm quite busy this week. Can't look into.

eymen-elkum:
I used it first time today, yes not daily basis at all, can be postponed or discarded.

PR #2727, opened 2023-05-03
Проверка версии

sab0tajue:
It is required to make a grouped report from two sections "Buildings" and "Lands". They are combined into one field "Building + Land" where the user attaches the building to the land. How can I display a grouped report for this related field so that you can see the related properties: Building + land?


sab0tajue:
sorry not there

PR #2730, opened 2023-05-05
Fix: kanban dashlet scrolling

SuchAFuriousDeath:
When using the kanban dashlet, the kanban header sticks itself to the top of the page, if you have dashlets above the kanban one.

The position method actually returns the position relative to its first non-static parent. In the classic list kanban, .position().top and .offset().top happened to return the same thing.

Hopefully this doesn't break anything. 
image

### DIFF
--- a/client/src/views/record/kanban.js
+++ b/client/src/views/record/kanban.js
@@ -399,9 +399,9 @@ define('views/record/kanban', ['views/record/list'], function (Dep) {
                     return;
                 }
 
-                let stickTop = this.$listKanban.position().top - topBarHeight;
+                let stickTop = this.$listKanban.offset().top - topBarHeight;
 
-                let edge = $middle.position().top + $middle.outerHeight(true);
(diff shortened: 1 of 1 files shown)


yurikuzn:
There's no Kanban dashlet in Espo.

yurikuzn:
I'll test when get a chance  on the next few days. This requires rigorous testing.

SuchAFuriousDeath:
> There's no Kanban dashlet in Espo.

Oh, I became so used to it that I thought it was in Espo. My apologies. Maybe you could add it once there's time, it's useful to display your tasks together on the main page together with other information.
What's also really nice is an 'advanced filters' field for the kanban and list dashlet that allows you to select filters just like in normal list view. I could do a pull request on both, if you're interested.

image

yurikuzn:
I applied the fix in the master.

Regarding new dashlet type, I will consider in a few months. As this moment, we are feature frozen till mid June.

PR #2734, opened 2023-05-09
Fixed count query with group by

mozkomor05:
Group by causes most of count queries to not work as result isn't properly aggregated. Every query with GROUP (independently of HAVING clause) needs wrapping into subquery.

This causes 1-N joins without implicit grouping to be unusable in appliers (as count query returns wrong count => pagination doesn't work).

Current workaround is to add always true having clause to queries, however I find this solution very hacky and inelegant.

### DIFF
--- a/application/Espo/ORM/Mapper/BaseMapper.php
+++ b/application/Espo/ORM/Mapper/BaseMapper.php
@@ -230,8 +230,7 @@ class BaseMapper implements RDBMapper
             ->build();
 
         $wrap = $aggregation === self::FUNC_COUNT && (
-            $select->isDistinct() ||
-            $select->getGroup() && $select->getHaving()
+            $select->isDistinct() || $select->getGroup()
         );
 
         if (!$wrap) {


yurikuzn:
> This causes 1-N joins without implicit grouping to be unusable in appliers (as count query returns wrong count => pagination doesn't work).

Could you elaborate more on it. I don't understand.

How to reproduce the problem?

yurikuzn:
Do you use having (without group by) for the list view? I don't think it will be a valid SQ,L, might produce an error in strict mode.

Maybe use IN sub-query.

yurikuzn:
I suspect this code was added when we used having with group-by in some place for the list view. I believe since it was fixed to use IN sub-query-with-group-by as group-by is eligible to be used for record list fetching.

mozkomor05:
> I suspect this code was added when we used having with group-by in some place for the list view. I believe since it was fixed to use IN sub-query-with-group-by as group-by is not eligible to be used for record list fetching and produces an error in MySQL by default since 5.7. I'd rather remove groupBy and having from the repository class I future.

IMO it's reasonable to use GROUP BY. As EspoCRM doesn't currently support selecting and joining subquery, it is not always possible to avoid it. I think COUNT should be exclusively counted via subquery, thus avoiding all aggregation problems.

yurikuzn:
>  doesn't currently support selecting and joining subquery

I believe that not-storable field + field loader will show similar performance as sub-query in select.

mozkomor05:
> I believe that not-storable field + field loader will show similar performance as sub-query in select.

But sometimes it's required to sort and search.

mozkomor05:
Implementing `joinSubquery` could solve these kind of problems without `GRUOP BY`. I understand that you are currently feature-closed and something like this would take a lot of testing.

yurikuzn:
I will take a look. This code might be not used the fix can be applied.

yurikuzn:
> Implementing joinSubquery could solve these kind of problems 

Could you provide an SQL example please.

mozkomor05:
```sql
SELECT ..., s.totalQuanttiy
FROM product
    INNER JOIN (
        SELECT id, SUM(warehouseItem.quantity) as 'totalQuantity'
        FROM product
            LEFT JOIN warehouse_item ON warehouse_item.product_id = id
        GROUP BY id
    ) AS s ON s.id = id

```

yurikuzn:
Created a PR adding the ability to join sub-queries. https://github.com/espocrm/espocrm/pull/2757

PR #2741, opened 2023-05-14
HACK to allow Portal access on custom address without FPM

### LINKED ISSUE #2740: Remove need for FPM in order to have Portals
The @tmachyshyn implementation via `docker-compose.yml` does not use `FPM` which then breaks the portal implementation:

* `docker-compose.yml`: https://docs.espocrm.com/administration/docker/installation/#install-espocrm-with-docker-compose
* `Portal nginx.conf`: https://docs.espocrm.com/administration/portal/nginx-configuration/

This was how I resolved it so far:

```php
//de HACK 
if (function_exists('apache_request_headers') ) {
        $headers = apache_request_headers();
        if (array_key_exists("Espo-Portal-ID", $headers)) {
                $_SERVER['ESPO_PORTAL_ID'] = apache_request_headers()['Espo-Portal-ID'];
                if ($_SERVER['ESPO_PORTAL_ID']) {
                        $_SERVER['ESPO_PORTAL_IS_CUSTOM_URL'] = true;
                }
        }
}
```

Right at the top of `public/portal/index.php` as in my fork:

https://github.com/MigratorONE/espocrm/commit/50e735a11cfb748654d5d840ca76461462c68123

Wondering where I ought to put that, or how best to do that, then update the documentation to include:

```
proxy_set_header Espo-Portal-ID $espocrm_portal;
```

Or similar: overall the `docker-compose` documentation has separate from the `FPM` implementation which the rest of the documentation for `EspoCRM` itself seems to use exclusively.

digitalextremist:
Resolves #2740 

But #2740 itself questions if this is the best way.

### DIFF
--- a/public/portal/index.php
+++ b/public/portal/index.php
@@ -27,6 +27,20 @@
  * these Appropriate Legal Notices must retain the display of the "EspoCRM" word.
  ************************************************************************/
 
+//de HACK
+//de Must pass Espo-Portal-ID header:
+//de i.e. nginx: proxy_set_header Espo-Portal-ID $espocrm_portal;
+
+if (function_exists('apache_request_headers') ) {
+        $headers = apache_request_headers();
+        if (array_key_exists("Espo-Portal-ID", $headers)) {
+                $_SERVER['ESPO_PORTAL_ID'] = apache_request_headers()['Espo-Portal-ID'];
+                if ($_SERVER['ESPO_PORTAL_ID']) {
+                        $_SERVER['ESPO_PORTAL_IS_CUSTOM_URL'] = true;
+                }
+        }
+}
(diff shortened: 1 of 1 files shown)


yurikuzn:
We try not to use such hucks.

digitalextremist:
No problem @yurikuzn: opening the discussion only, with proof of concept. The @tmachyshyn implementation of `docker-compose.yml` does not work with the FPM way of distinguishing custom portal URLs. Just need to do it right. For me, I pass the portal header on an `nginx` level, at the reverse proxy. Nothing can change it after that, and if it is already set, it is cleared. But yes, without that protection, a header could be passed to the server and change the portal.

PR #2743, opened 2023-05-18
Attributes option on modal edit

eymen-elkum:
This option `attributes` only considered when model is new, expected to be addressed in edit mode as well.


yurikuzn:
I'm not sure about this fix. As we have tested v7.5 before upcoming release I can't apply it w/o having a reproducible bug.

Unsetting object passed to a view is not acceptable w/o cloning. I'm extremely busy these weeks, can't look into.

eymen-elkum:
This is actually not an issue because Espo doesn't pass attributes when editing. I needed this functionality for custom development and discovered that it might be useful and beneficial.

The scenario occurs when the model has already been modified but not yet submitted in the list view, and we call 'editQuick'.

eymen-elkum:
Please feel free to postpone or cancel this merge. I am going to save the row model before the modal edit as a workaround.

yurikuzn:
I'm still not sure about this addition. I'd just write something like this:

```js
const view = new EditModalView({...});
await this.assignView('dialog', view);

view.model.set(attributes);

await view.render();
```

PR #2812, opened 2023-08-01
Localization of templates (de_DE)

AurelioSilver:
Translated all templates to german.


yurikuzn:
Thanks for contributing. I'll review a bit later.

PR #2813, opened 2023-08-01
Localization of templates (it_IT)

Kharg:
Translated all templates to Italian.


yurikuzn:
It seems your git somehow does not honor https://github.com/espocrm/espocrm/blob/7.5.5/.gitattributes#L8

When I merge your commits, git applies changes (unstaged) fixing new lines.

Kharg:
I am sorry for that.
I used VSCode to translate the templates. I will check both Git and VSCode settings.

Do I need to make a new pull request?

yurikuzn:
IFIK VSCode does not honor editorconfig out of the box. https://github.com/espocrm/espocrm/blob/7.5.5/.editorconfig#L7. There's is a plugin. But Git should fix new lines when you `git add`. Somehow it does not, maybe you pass a parameter to ignore it.

PR #2834, opened 2023-09-08
fix version check

eymen-elkum:
Clients were able to upgrade their crm even though it should not .  

In my extension I set this:

    "acceptableVersions": [
        ">=7.0.8",
        "<=8.0.0"
    ],

### DIFF
--- a/application/Espo/Core/Upgrades/Actions/Base.php
+++ b/application/Espo/Core/Upgrades/Actions/Base.php
@@ -428,19 +428,19 @@ abstract class Base
                 $this->getLog()->error('SemVer: Version identification error: '.$e->getMessage().'.');
             }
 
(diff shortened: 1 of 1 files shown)


arkadywtf:
I think it's supposed to be a range, so this will work: ">=7.0.8 <=8.0.0" which we use in our extensions.

yurikuzn:
Anyway we can't make changes in the hotfix branch in such a vital functionality that could break the upgrade flow. Might need to fix.

Please, more info in PRs and issues. It's hard to switch context, especially so little input is provided.

yurikuzn:
For such problems, it's better to create an issue first. arkadywtf is right I believe and this PR would break the correct logic.

PR #2835, opened 2023-09-09
language and null checking updates

bandtank:
The language updates are minor. Some of the phrases have been changed to sound more natural.

Regarding the phone number issue, I was seeing this in the log:
```
E_DEPRECATED: trim(): Passing null to parameter #1 ($string) of type string is deprecated {"code":8192,"message":"trim(): Passing null to parameter #1 ($string) of type string is deprecated","file":" .../application/Espo/Core/FieldProcessing/PhoneNumber/Saver.php","line":414}
```

It happened here:
```
      $user->set("teamsIds", $teamIds);
      $this->em->saveEntity($user);
```

Saving a user entity that does not have a phone number caused the warning.


yurikuzn:
Better to have different issues in separate PRs.

`?? ""` fix is not good I think, I will look into to come up with something more appropriate.

The language fixes in en_US is a pain to deal with. As any change in en_US breaks all translations in PO. The current design has flaws.

application/Espo/Resources/i18n/en_US/Global.json
```diff
@@ -305,7 +305,7 @@
         "fieldNotMatchingPattern$uriOptionalProtocol": "{field} must be a valid URL",
         "fieldInvalid": "{field} is invalid",
         "fieldIsRequired": "{field} is required",
-        "fieldPhoneInvalidCharacters": "Only digits, latin letters and characters `-+_@:#().` are allowed",
+        "fieldPhoneInvalidCharacters": "Only digits, latin letters, and characters `-+_@:#().` are allowed",
```
yurikuzn, line 308: > The Oxford Companion to the English Language notes: "Usage varies as to the inclusion of a comma before *and* in the last item. ... This practice is controversial and is known as the serial comma or Oxford comma, because it is part of the house style of [Oxford University Press](https://en.wikipedia.org/wiki/Oxford_University_Press)."[[15]](https://en.wikipedia.org/wiki/Serial_comma#cite_note-15)

bandtank:
That makes sense. I will keep that in mind in the future.

PR #2840, opened 2023-09-11
fix: The `prepare` method may cause a rendering problem in rare situations

### LINKED ISSUE #2839: The `prepare` method may cause a rendering problem in rare situations
Sometimes it is possible to get into situations where the `prepare` method is called, but is not followed by a re-render. This will cause particular problems if a nested view is created in `prepare` (as is the case with the time-ranges field, for example).

**Theoretical reproduction**
1. Have a client-read-only field that changes value during record editing (e.g. due to dyanmic handler)
2. Cancel editing
3. The following code undoes the changes to the model, but skips the re-render. 
https://github.com/espocrm/espocrm/blob/a20c3f14ace3d9ea266d52240ec32415c5724b9c/client/src/views/record/detail.js#L1656
4. As a result, only `prepare` is called, without subsequent re-rendering.

**Example**
1. Create a read-only test field with the following custom view: 
```js
define(["views/fields/varchar"], Dep => {
    return class extends Dep {
        readOnly = true;

        _template = '<div class="field">{{{myView}}}</div>';

        setup() {
            super.setup();

            // for the sake of simple test case
            window.triggerChange = () => {
                this.model.set('test', Math.random().toString(36).substring(2, 15));
            };
        }

        prepare() {
            return this.createView('myView', 'views/base', {
                templateContent: "Hello!",
                selector: '.field',
            });
        }
    };
});
```
4. Add it to the layout.
5. Create a new record and edit it.
image
image

6. Call `window.triggerChange()` (used to simulate a real field value change during edit mode, can occur e.g. by dynamic handler)
image
7. Cancel the editation.
7. **Nested view `myView` will not render.** 
image

**Proposed solution**
Do not call `prepare` in situations where it is unconditionally known in advance that a render cannot occur. This solution, in my opinion, is more general and may prevent more problems with `prepare` in the future.

mozkomor05:
Closes #2839

### DIFF
--- a/client/src/views/fields/base.js
+++ b/client/src/views/fields/base.js
@@ -865,18 +865,20 @@ class BaseFieldView extends View {
                     return;
                 }
 
+                if (options.skipReRenderInEditMode && this.isEditMode()) {
+                    return;
+                }
+
+                if (options.skipReRender) {
+                    return;
+                }
+
                 let reRender = () => {
                     if (!this.isRendered() && !this.isBeingRendered()) {
                         return;
                     }
 
-                    if (options.skipReRenderInEditMode && this.isEditMode()) {
-                        return;
-                    }
-
-                    if (!options.skipReRender) {
-                        this.reRender();
-                    }
+                    this.reRender();
 
                     if (options.highlight) {
                         this.trigger('highlight');


yurikuzn:
Thanks. The fix looks proper. Anyway, since we (almost) don't use `prepare` I thinks it's safe to merge it into fix.

PR #2853, opened 2023-09-19
German Translation fixes and improvements

AurelioSilver:
This adds translated cancellation templates and fixes a few spelling errors or improves the translation.

The translations are also updated on poeeditor.com. 
Only the most serious ones are included in this PR.


yurikuzn:
Thanks for contribution. We don't apply translations (except templates) via PRs. We build from PO Editor then merge into repository.

I applied the latest changes from PO into the fix branch. GitHub didn't adjust the diff for some reason.

AurelioSilver:
Ok, should I create a new PR with only the templates? Or remove the PO from this pull request (not sure how to do it, hard reset to the first commit?)

Issue #2330, opened 2022-05-30
Shared calendar views in week mode have different UI elements


bandtank:
**Is your feature request related to a problem? Please describe.**
In the calendar, the week view for the logged in user has separators at 15-minute increments. The events also display the beginning and end times. Both features are useful for a variety of reasons. However, in shared calendars, the separators are missing and only beginning times are displayed. 

**Describe the solution you'd like**
I would like both calendar views to look exactly the same. It makes sense for the views to be consistent; I am not sure why the shared view has fewer features.

**Additional context**
Here is how the personal calendar looks in week view:
![image](https://user-images.githubusercontent.com/1170415/171066643-f4bb0fbc-8f99-4e75-8698-e7a541882ead.png)

Here is how a shared calendar looks in week view:
![image](https://user-images.githubusercontent.com/1170415/171066628-b1c3d4b7-9a34-4db3-b397-59834e0bad84.png)

yurikuzn:
There would be multiple events overlapping on the same timeslots. That's why it was not implemented.

bandtank:
That's the issue, though. There are often times when multiple events are scheduled during the same time slot even for a single user. Here is how Outlook handles it for a single person:

<img width="567" alt="image" src="https://user-images.githubusercontent.com/1170415/171152210-13ce1921-ef7e-48b8-9a52-4c4d8f5df831.png">

and for multiple people:

<img width="302" alt="image" src="https://user-images.githubusercontent.com/1170415/171152455-666fdd6a-1eef-460d-b3b2-21a50d7e0e2d.png">

Espo overlaps the events in single-user mode, which is difficult to visually understand:

<img width="466" alt="image" src="https://user-images.githubusercontent.com/1170415/171152747-0e04c336-a9de-4499-b9b4-a0b0886e0d0b.png">

When multiple users are added, we have a hard time seeing what is happening because too many events stack on top of each other and the gaps can be confusing when many attendees are added to a meeting (the blue meeting, for example):

![image](https://user-images.githubusercontent.com/1170415/171154487-d960c9e7-0458-4397-b79e-d6d22c73e8f5.png)

Making the views the same would fix the issue. Widening the column by adding a horizontal scroll bar would allow events to overlap in time without overlapping visually. Is there a way to override this behavior with a custom view?

yurikuzn:
We use  FullCalendar lib, I can't make any changes there. It would add a maintenance burden.

bandtank:
[...]

PR #2862, opened 2023-09-27
Add shared calendar options for agendaWeek and agendaDay

bandtank:
Adding these options allows users to create shared views that align with other software, such as Outlook, which allows time-based views of multiple users' calendars. The overlapping events were discussed in another issue (#2330) in [this comment](https://github.com/espocrm/espocrm/issues/2330#issuecomment-1141698632), which is understandable for some use cases, but overlapping events are often desirable to see. This change allows users to have the flexibility to choose without incurring maintenance costs or additional work as `FullCalendar` already supports both modes.

The addition of the modes results in two new options in the dropdown menu when creating a shared calendar:

image


yurikuzn:
Can't apply this change.

Issue #2860, opened 2023-09-27
Calendar: Wrong week displayed when refreshing or using date query parameter


bandtank:
**Describe the bug**
Refreshing the page with the browser's refresh button when viewing the calendar in Week mode (single user and shared) causes the calendar to display the wrong week sometimes. The issue appears to be related to the `date` query parameter. The bug can also be seen if you enter a date directly into the URL as a query parameter, e.g. `&date=2023-09-10`.

**To Reproduce (MANDATORY, DO NOT REMOVE)**
#### Pre-requisites:
Set the application's timezone to `America/Denver` and first day of the week to `Sunday`.

#### Reproducing the bug with the refresh button:
1. Go to the calendar
2. Set it to week mode
3. Scroll to September 17, which is a Sunday, using the arrow buttons at the top left of the calendar view.
4. As you scroll, be sure the `date` query parameter is updating to match the first day of the week. For example, when you scroll to the right to set the first column as September 17, the query parameter should be `&date=2023-09-17`. The page should look like this:
![image](https://github.com/espocrm/espocrm/assets/1170415/45c7ae26-fdd1-4cda-a03e-792dbfe54390)
and the URL should look like this: `http://espocrm-fork:8081/#Calendar/show/mode=agendaWeek&date=2023-09-17`
6. Scroll to the left by 1 week using the arrow. The query parameter should now be `&date=2023-09-10`. The page should look like this:
![image](https://github.com/espocrm/espocrm/assets/1170415/aba02d41-af0f-499b-8a69-7dd6c7de1d54)
and the URL should look like this: `http://espocrm-fork:8081/#Calendar/show/mode=agendaWeek&date=2023-09-10`
7. Refresh the page using the browser's refresh button. The calendar now looks like this:
![image](https://github.com/espocrm/espocrm/assets/1170415/ea6f885f-90f3-42f7-81b9-f67626b9ae51)
and the URL still looks like this: `http://espocrm-fork:8081/#Calendar/show/mode=agendaWeek&date=2023-09-10`

#### Reproducing the bug with the query parameter:
Instead of scrolling through the calendar, go directly to the first day of a week using the query parameter. In my case, this only happens when the date is a Sunday because the first day of the week is configured as Sunday. Entering a date that is a Saturday or Monday will cause the week to render correctly.

Entering this URL: `http://espocrm-fork:8081/#Calendar/show/mode=agendaWeek&date=2023-09-10`
Results in this view:
[...]

PR #2864, opened 2023-09-27
Change the FullCalendar initialDate parameter to local instead of UTC

bandtank:
I believe this is the fix for #2860.


yurikuzn:
Applied separately to the fix branch: https://github.com/espocrm/espocrm/commit/8bce12aca3f50fef339b7d8c9ecb24f7b13d5129

PR #2866, opened 2023-09-28
fix varchar autocomplete

eymen-elkum:
This fix case when you select will never stop calling the autocomplete ajax requests again and again.

You may have better way, but didn’t manage to find better.


eymen-elkum:
tried to reproduce on demo not managed, maybe because of customisation on my side

PR #2881, opened 2023-10-23
Incorrect metadata path for "validatorClassNameMap"

DrWarpMan:
The 'validatorClassNameMap' metadata is incorrectly read from `fields.<fieldName>`.
The correct path should be: `fields.<fieldType>`.

https://github.com/espocrm/espocrm/blob/bcc6edf083567b099a791ec162a79f97cf450c56/application/Espo/Core/FieldValidation/ValidatorFactory.php#L72


yurikuzn:
It's worth adding `$fieldType ?? ''` to prevent null passed for `string[]`, Phpstan on level 8 does not detect such issues when type is mixed. I'll do it.

PR #2905, opened 2023-11-20
Add Export functionality to Import detail page panels

bhbrunt:
### Background
On the Import detail page there are panels showing the imported, updated and duplicate records associated with that import. It's useful to be able to export these lists for verification purposes. 

Export of these lists can currently be (partially) achieved by opening the "View List" modal, clicking the "Select all" checkbox and selecting Export on the dropdown. This has some shortcomings, however:
- Only the records currently displayed are included in the export, even if many more are present. It is (I think) not possible to export the entire list, and it is easy to mistakenly think the export includes all records.
- The export option is accessible only from the View List modal, and not directly from the detail page where the panels are displayed.

There is existing functionality to export the list of import errors from the Import detail page, but this is not available for the relationship panels (imported, updated, duplicates).

### Description

This pull request implements a CSV export of the records linked to an import.
This adds an "Export" item to the dropdown menu for each panel on the import detail view, in the same style as exists for the import errors list.
The implementation follows the existing functionality for exporting the list of "error" records, and uses the existing CSV export functionality.

image

### DIFF
--- a/application/Espo/Resources/i18n/en_US/Import.json
+++ b/application/Espo/Resources/i18n/en_US/Import.json
@@ -61,6 +61,7 @@
     "messages": {
         "importRunning": "Import running...",
         "noErrors": "No errors.",
+        "noRecords": "No records were found to export.",
         "utf8": "Should be UTF-8 encoded",
         "duplicatesRemoved": "Duplicates removed",
         "inIdle": "Execute in idle (for big data; via cron)",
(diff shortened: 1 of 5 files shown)


yurikuzn:
Thanks for PR. We are currently in a process of creating a new version of our CLA, it's going to be available in a few days.

I'll check this more detail when get a chance. Not yet sure whether we need this functionality though.

A few things I noticed while viewing quick:

* Espo\Services\Import is not right choice.
* Indentation problem in some places. Might be your editor/IDE does not honor editorconfig and uses tabs instead spaces.

bhbrunt:
Thanks for the first review! I've fixed the indentation problems, now taking account of editorconfig and 4 spaces rather than tabs.

Issue #2919, opened 2023-12-08
S3 Compatible Storage


arkadiyasuratov:
**Is your feature request related to a problem? Please describe.**
The AsyncAws client supports non-AWS S3 endpoints, and it would be great if we could use this with an existing file storage implementation.

**Describe the solution you'd like**
That is why I'm creating an issue first, as this can go in two ways.

**Proposal 1**

```php
  'awsS3Storage' => [
    'bucketName' => 'BUCKET_NAME',
    'path' => 'OPTIONAL_PATH_WHERE_FILES_WILL_BE_STORED',
    'credentials' => [
      'key' => 'KEY',
      'secret' => 'SECRET',
    ],
    'region' => 'YOUR_REGION',
    # Any options that can be passed to the AsyncAws client
    # https://async-aws.com/configuration.html#configuration-reference
    'clientOptions' => [
      'pathStyleEndpoint' => true,  # Needed for backends such as MinIO, DigitalOcean Spaces, etc.
      'endpoint' => 'https://eu.r2.cloudflarestorage.com',
      // 'endpoint' => 'https://fra1.digitaloceanspaces.com',
      // 'endpoint' => 'https://play.min.io',
    ],
  ],
```

**Proposal 2**

```php
  'awsS3Storage' => [
    'bucketName' => 'BUCKET_NAME',
    'path' => 'OPTIONAL_PATH_WHERE_FILES_WILL_BE_STORED',
    'credentials' => [
      'key' => 'KEY',
      'secret' => 'SECRET',
    ],
    'region' => 'YOUR_REGION',
    'pathStyleEndpoint' => true,   # Needed for backends such as MinIO, DigitalOcean Spaces, etc.
    'endpoint' => 'https://eu.r2.cloudflarestorage.com',
    // 'endpoint' => 'https://fra1.digitaloceanspaces.com',
    // 'endpoint' => 'https://play.min.io',
  ],
```

**Describe alternatives you've considered**
Currently, I'm using a forked version of the `Espo\Core\FileStorage\Storages\AwsS3` class, which is not ideal.

```php
<?php

namespace Espo\Modules\DubasCrm\Core\FileStorage\Storages;

use AsyncAws\S3\S3Client;
use Espo\Core\FileStorage\Attachment;
use Espo\Core\FileStorage\Storage;
use Espo\Core\Utils\Config;
use GuzzleHttp\Psr7\Stream;
[...]

PR #2920, opened 2023-12-10
Chunked body for Non-AWS S3 endpoints

arkadiyasuratov:
Related to https://github.com/espocrm/espocrm/issues/2919

When testing Cloudflare R2 endpoint with larger files I encountered the following error:

```
ERROR: (501) HTTP 501 returned for "eu.r2.cloudflarestorage.com/6761da60-e9c2-4794-9303-f404f92360ea".
Code:    NotImplemented
Message: STREAMING-AWS4-HMAC-SHA256-PAYLOAD not implemented
Type:
Detail:   ; POST /Attachment; line: 402, file: /var/www/html/vendor/async-aws/core/src/Response.php
```

To fix this I needed to set `sendChunkedBody` as `false`.

Regarding this line: `$sendChunkedBody = $config->get('awsS3Storage.sendChunkedBody') ?? null;`

The reason `sendChunkedBody` defaults to `null` is because the documentation for AsyncAws is unclear. Here, it suggests that the default value is `false`: https://async-aws.com/configuration.html#sendchunkedbody. However, here, it seems to suggest that it is enabled by default: https://async-aws.com/clients/s3.html#chunked-body. To avoid confusion, it might be better not to pass this parameter to the client by default, or to have a clearer name such as `sendChunkedBodyDisabled`.

### DIFF
--- a/application/Espo/Core/FileStorage/Storages/AwsS3.php
+++ b/application/Espo/Core/FileStorage/Storages/AwsS3.php
@@ -57,6 +57,7 @@ class AwsS3 implements Storage
         $credentials = $config->get('awsS3Storage.credentials') ?? null;
         $endpoint = $config->get('awsS3Storage.endpoint') ?? null;
         $pathStyleEndpoint = $config->get('awsS3Storage.pathStyleEndpoint') ?? false;
+        $sendChunkedBody = $config->get('awsS3Storage.sendChunkedBody') ?? null;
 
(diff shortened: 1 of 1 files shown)


yurikuzn:
Nice.

The parameter seems is defaulted to false, according this: https://github.com/async-aws/aws/blob/master/src/Service/S3/src/Signer/SignerV4ForS3.php#L55

PR #2953, opened 2024-01-23
Fix for long messages in the logger causing a runtime exception

mozkomor05:
Incorrect use of `with` method causes long error messages to fail with RuntimeException. 

```
Error message: Named parameter $channel overwrites previous argument; at /home/******/public_html/application/Espo/Core/Log/Handler/EspoFileHandler.php:88.
```

https://github.com/Seldaek/monolog/blob/479c936d2c230d8c467bdb3882afab45a6e6b8ad/src/Monolog/LogRecord.php#L116-L122

image

Explanation: incorrect usage of `with` method causes `$args[1]` to be set to string. Constructor parameter#1 also happens to be `channel` with is also present in the `$args` array. `$args` is then used to initialize the instance of new LogRecord. This causes named parameter `$channel` to overwrite previous argument on the position `#1`.

### DIFF
--- a/application/Espo/Core/Log/Handler/EspoFileHandler.php
+++ b/application/Espo/Core/Log/Handler/EspoFileHandler.php
@@ -97,7 +97,7 @@ class EspoFileHandler extends MonologStreamHandler
 
         $message = substr($record->message, 0, $this->maxErrorMessageLength) . '...';
 
-        $record = $record->with('message', $message);
+        $record = $record->with(message: $message);
 
         return $this->getFormatter()->format($record);
     }


mozkomor05:
Note: this should probably also be fixed in the `fix` branch.

radek3911:
+1 since upgrading to 8.1.1 this has became issue for long logs (for debug purposes I log large data from external systems, while not best practice it's mainly for audit and helps with troubleshooting), if the log got too long EspoFileHandler would throw exception on LN 88 and kill execution of any job/API action that logged long line.
image

yurikuzn:
Thanks for reporting. Bad that the static analysis didn't catch the error, as there not strict enough type in the lib. I applied the fix to the fix branch.

yurikuzn:
I hope one day PHP arrays will deprecated or just forgotten. But it seems will never happen.

mozkomor05:
I know right. Set? Hash table? Dynamic arrays? DTOs? Everything is just `array` which creates room for many mistakes :) I like Symfony's preference for DTOs instead of arrays.

yurikuzn:
Luckily, arrays are not used so often for data transferring nowadays. But problem is with an array as a data structure, as it's both a hash-table and list. More, a hash table that can be interpreted as list in some cases.

It would be a good start introducing two separate types: list/array/vector and dictionary/hash. Like what Facebook did to PHP. For some reason there's no any movement to that direction though almost anyone would agree that the current arrays are problematic.

Some of issues that happen a lot:

1. A list after filtering may become a hash.
2. json_encode can't know whether an empty array is a hash or a list. We had a big problem with this in early days in Espo, when metadata was sent to the frontend, the code couldn't know whether a particular empty array was supposed to be an object or a list.

yurikuzn:
Also, it would be great if there were Array and Dictionary classes in STD lib.

Writing this

```php
return $list
    ->filter(fn ($it) => $it->isActive())
    ->map(fn ($it) => $it->getId());    
```

takes much less time and effort than

```php
$list = array_filter(// can't remember argument order, can't write w/o looking into docs);
$list = array_map(// the same);
return array_values($list); // can't be forgotten
```

yurikuzn:
There was a good RFC https://wiki.php.net/rfc/vector, though intention was more of having separate types vector and deque rather than movement into abandoning PHP arrays.

mozkomor05:
Abandoning arrays would be a really big step for PHP. Similarly, introducing a more robust type system would be fantastic, though we can already see some progress there.
Personally, I think that adding data structures is a great first step in the right direction.

Issue #2284, opened 2022-04-20
Add a filter for This Week


bandtank:
In reports, formulas, etc., there are filters for today, this month, this quarter, and this year. I often need to find records in a specific week, which is difficult without built-in support. I realize it’s not as straightforward to select records in a specific week because a week can span different days; multiple options could exist or the config parameter could be used.

Being able to find records with a datetime field that is >= 00:00 on Sunday and < 00:00 on the next Sunday would be very useful.

bandtank:
It's not quite the same as what I had previously requested, but the following functions provide a `This Week` filter. I would eventually like to add this filter to reports as a default option.

```
<?php

namespace Espo\Custom\Classes\Select\Holiday\PrimaryFilters;

use Espo\Core\Select\Primary\Filter;
use Espo\Core\Select\Helpers\UserTimeZoneProvider;
use Espo\Custom\Services\Common;
use Espo\ORM\Query\SelectBuilder;

use DateTimeZone;
use DateTime;

class ThisWeek implements Filter
{
    public function __construct(
        private UserTimeZoneProvider $userTimeZoneProvider,
        private Common $common
    ) {}

    public function apply(SelectBuilder $queryBuilder): void
    {
        $tzLocal = new DateTimeZone($this->userTimeZoneProvider->get());
        $anchor = (new DateTime("now", $tzLocal));

        list($dateStart, $dateEnd) = $this->common->getWeekStartEndUtc($anchor);

        $where = array(
          "dateStart<" => $dateEnd->format("Y-m-d H:i:s"),
          "dateEnd>=" => $dateStart->format("Y-m-d H:i:s"),
        );
        $queryBuilder->where($where);
    }
}
```

```
  public function getWeekStartEndUtc($anchor = Null) {
    if($anchor === Null) {
      $tzLocal = new DateTimeZone($this->config->get('timeZone', 'UTC'));
      $anchor  = (new DateTime())->setTimezone($tzLocal);
    }

    $tzUtc = new DateTimeZone('UTC');
    $anchor->setTime(0,0)->setTimezone($tzUtc);

    $begin = (clone $anchor);
    $end   = (clone $anchor);

    $dayOfWeek = intval($begin->format('w'));
    $begin->modify("-" . $dayOfWeek . " days");
    $end->modify("+" . (7 - $dayOfWeek) . " days");

    return [$begin, $end];
  }
```

For reports, the following expression works for half of the year. It is off by 1 hour for the other half of the year:
[...]

PR #2965, opened 2024-01-28
Filters for date and datetime to search by week

bandtank:
This request adds a few filters to date and datetime fields:
* Current Week - search for records within the current week (based on the user's weekStart preference or the system's weekStart setting if the user does not have a preference)
* Last Week - Look for records in the last full week (the current week is not searched)
* Last X Weeks - Look for records in the last X full weeks (the current week is not searched)
* Next Week - Look for records in the next full week (the current week is not searched)
* Next X Weeks - Look for records in the next X full weeks (the current week is not searched)

Many businesses operate on a weekly basis, which makes searching, reporting, and filtering records with weekly filters very attractive. I currently build these filters into my instances manually. Instead of continuing to do that, I took the time to add the features to the application in a (hopefully) maintainable way.

If you do not like this feature or some of the changes, please let me know if there is anything I could change for it to be accepted. Otherwise, I will probably release the features as an extension, presuming I can figure how to do that.

image

image


bandtank:
#2284

yurikuzn:
Hi,

This raise some design issues. It was supposed that the Converter class is user agnostic. I'll consider this enhancement when get a chance, maybe will re-implement, not sure.

To add filters in a module:

* Define custom where item converters: https://docs.espocrm.com/development/metadata/app-select/#whereitemconverterclassnamemap
* Use monkey patching to add additional search types for date field view: https://forum.espocrm.com/forum/developer-help/100485-how-to-add-javascript-code
* For DateTimeTranformer, it may be trickier.

bandtank:
Ok, sounds good. I thought about adding the feature in a module, but I couldn't figure out how to work around DateTimeTranformer. I'll see if I can figure something out. Thanks

yurikuzn:
See https://github.com/espocrm/espocrm/blob/8.1.1/application/Espo/Core/Select/Where/ConverterFactory.php#L150. As an option, you can have this filters only for date view.

yurikuzn:
You can also by pass the date-time transformer and rely only on where item converter for both date-times and dates. I think this would be the easiest and future-proof.

bandtank:
I am closing this for now because the commits are causing too much noise. I will reopen the PR when the work is done and ready to be reviewed. Hopefully the new approach works betters.

PR #3008, opened 2024-03-06
Map provider config paramater

rabiibrahimi:
Implmenting a custom map provider is a great option, however there is no way for the user to choose which provider to apply to the maps fields in the system and therefore the system would always apply Google as default config, adding a map provider config params to the UI would allow the user to choose which map provider to use.


bandtank:
Please undo the whitespace changes. The diffs are much more significant than they should be.

Kharg:
Hey @yurikuzn any plans on merging this?

I am developing an open street map extension and I am planning to release it for free. Having a field to choose the provider natively would be a great addition!

yurikuzn:
There's a probability that the map functionality will be revamp in the future.

PR #3013, opened 2024-03-14
Print extension version in console using command.php

SinyaWeo:
Along with extension name prints extension version when using 'php command.php extensions --list' command.

Useful for automatic server data management.


yurikuzn:
Added: https://github.com/espocrm/espocrm/commit/223b4ab8aa16d91d180c0f7b2599410d67ffa9c4

PR #3026, opened 2024-04-04
add missing & to google static url

eymen-elkum:


eymen-elkum:
Not sure but maybe need also to remove the ; from the format 

```
-        $format = 'jpg';
+        $format = 'jpg;';


```

PR #3107, opened 2024-07-22
fix: Resolve error 500 with Microsoft Office 365 OIDC authentication …

Gebetshaus:
Solves the problem of error 500 when using Microsoft office 365 authentication (OIDC) with two-factor authentication enabled in Espocrm. 
Currently, a ZFA code query is also executed with OIDC authentication and enabled ZFA in Espocrm.  
Goal: Internal authentication in Espocrm and authentication with external authentication providers (OIDC) should be separated. Either the authentication is carried out with the internal authentication, which queries the zfa code when two-factor authentication is activated in Espocrm, or the authentication is carried out with an OIDC provider that uses its own two-factor methods, in which case no ZFA code query from EspoCrm is necessary.

### DIFF
--- a/application/Espo/Core/Authentication/Authentication.php
+++ b/application/Espo/Core/Authentication/Authentication.php
@@ -236,7 +236,8 @@ class Authentication
         if (
             !$result->isSecondStepRequired() &&
             !$authToken &&
-            $this->configDataProvider->isTwoFactorEnabled()
+            $this->configDataProvider->isTwoFactorEnabled() &&
+            $username !== '**oidc'
         ) {
             $result = $this->processTwoFactor($result, $request);
 


yurikuzn:
I'll try to find a non hacky solution when get a chance.

Gebetshaus:
Thank you!

Von: Yurii Kuznietsov ***@***.***>
Gesendet: Montag, 22. Juli 2024 13:04
An: espocrm/espocrm ***@***.***>
Cc: IT ***@***.***>; Author ***@***.***>
Betreff: Re: [espocrm/espocrm] fix: Resolve error 500 with Microsoft Office 365 OIDC authentication … (PR #3107)


I'll try to find a non hacky solution when get a chance.

yurikuzn:
Here's the solution: https://github.com/espocrm/espocrm/commit/a109e1353d8e91eeadb049401cc6237250b6d26f

Not yet tested.

PR #3115, opened 2024-08-08
Lower case and trim keywords in admin index quick search

eymen-elkum:
This is helpful when translation written with spaces or first capital letter


yurikuzn:
There are empty space on this line which resulted in a merge conflict: https://github.com/eymen-elkum/espocrm/blob/1f050b2763bc43b5f2405d2f18f03818519e569a/client/src/views/admin/index.js#L87

Looks like the editor you used does not honor this parameter: https://github.com/espocrm/espocrm/blob/master/.editorconfig#L9

eymen-elkum:
I will double checking formatting next time 👍🏼

This is because I made it directly online this time

PR #3123, opened 2024-08-24
User preferences helper

bandtank:
This helper allows developers to search a user's preferences by key. It is very similar to the UserTimeZoneProvider helper.


bandtank:
I'm moving this to a module.

PR #3131, opened 2024-08-27
Add DateTime functions and utilities for managing weeks

bandtank:
This branch introduces three functions:
* `addWeeks` in `application/Espo/Core/Field/DateTime.php`. The function is similar to `addMonths`, except key word is `weeks` instead of `months` in the format string. This allows a developer to `$date->addWeeks(-2)`;
* `getWeekStart` in `application/Espo/Core/Utils/DateTime.php`. The function allows developers to get a datetime object for the beginning of any week. With no parameters, the week start is 0, the target date is now, and the timezone is the system default.
* `getWeekStartEnd` in `application/Espo/Core/Utils/DateTime.php`. The function does the same thing as the previous function for the week start and it includes a datetime object for the week end as well.

This is useful because developers can quickly and easily get objects for the beginning and end of any week of the year:
```php
$days = $this->dateTimeUtil->getWeekStartEnd();
```

If you think these features are useful, but you need changes to be made, please let me know. I would be happy to do whatever is necessary because I frequently need to modify dates and datetimes in terms of weeks. Native support would dramatically simplify the process in many situations. I think it also makes sense to include support for weeks because the system manages days, months, quarters, and years.

---
Here are a few examples based on the current day (2024-08-27):

## All default parameters:
```php
$date = $this->dateTimeUtil->getWeekStart();
$GLOBALS["log"]->debug(
  "start of week with weekStart = 0",
  [$date->toDateTime()->format(DateTimeUtil::SYSTEM_DATE_TIME_FORMAT)]
);
```
Output:
```
DEBUG: start of week with weekStart = 0 ["2024-08-25 00:00:00"]
```

## Changing the weekStart to 4 (Thursday):
```php
$date = $this->dateTimeUtil->getWeekStart(4);
$GLOBALS["log"]->debug(
  "start of week with weekStart = 4",
  [$date->toDateTime()->format(DateTimeUtil::SYSTEM_DATE_TIME_FORMAT)]
);
```
Output:
```
DEBUG: start of week with weekStart = 0 ["2024-08-25 00:00:00"]
```

## Using a custom date:
```php
$date = DateTime::fromString("2024-08-10 11:01:02");
$dates = $this->dateTimeUtil->getWeekStartEnd(0, $date->toDateTime());
$GLOBALS["log"]->debug(
  "start of week with weekStart = 0 on 2024-08-10",
  [$dates["start"]->toDateTime()->format(DateTimeUtil::SYSTEM_DATE_TIME_FORMAT)]
);
$GLOBALS["log"]->debug(
  "  end of week with weekStart = 0 on 2024-08-10",
  [$dates["end"]->toDateTime()->format(DateTimeUtil::SYSTEM_DATE_TIME_FORMAT)]
);
```
Output:
```
DEBUG: start of week with weekStart = 0 on 2024-08-10 ["2024-08-04 00:00:00"]
DEBUG:   end of week with weekStart = 0 on 2024-08-10 ["2024-08-10 23:59:59"]
```


yurikuzn:
Hi,

Is  `addWeeks(n)` the same as `addDays(n * 7)` ?

Regarding methods in DateTimeUtil. I'm not a fan of adding anything to that class. This class is rather a legacy and has procedural-style methods what I try to avoid nowadays. Also I tend to see any code that is not used in the Espo codebase as a candidate for a removal. Adding helpers which are not used in the codebase is not what I would like in the project. We would need to maintain them, we would need to write tests for them. It's like having a library in the project.

We'd rather come up with some simple ability to use the same composer library by multiple different modules installed on the same instance. Now it's possible, but I believe it's not neat.

bandtank:
> Is addWeeks(n) the same as addDays(n * 7) ?

Yes. I don't think that function is necessary, but I figured I'd mention it just in case.

Regarding the other comments, I understand what you mean. After this PR, I was going to submit a PR to add week-based filters for the Espo field types called `Date` and `Date-Time`, which could use the `getWeekStart` and `getWeekStartEnd` functions. If you are not interested in those types of filters, I will create a module instead.

The reason I considered a PR is because the built-in views that use `date.js` and `datetime.js` apply to many fields, such as createdAt and modifiedAt, which would be nice to be able to filter by `Current Week` or similar filters. I think a module can override that behavior by modifying metadata, but I was trying to avoid messing with metadata. By that I mean figuring out a way to override things like this:
```
"view": "crm:views/meeting/fields/date-end",
```
Other than changing the `view` to point to a replacement view, I couldn't figure out how to add the filters to date-end.

bandtank:
I just saw your edit. I will try to make use of Carbon instead.

PR #3132, opened 2024-08-28
Get only active users in email address lookup

traien:


yurikuzn:
There's no *active* field. There's *isActive*.

traien:
> There's no _active_ field. There's _isActive_.

My bad, fixed it with new commit, or do I have to create another PR?

Issue #3110, opened 2024-07-27
Add a separate config parameter for logging failed sql 


SuchAFuriousDeath:
**Is your feature request related to a problem? Please describe.**
It is very useful to know a full failed sql query for debugging purposes. However, it is only possible to log failed sql by turning on the logging of all sql, which is very inconvenient (It makes the log very big and it makes it very difficult to watch for new changes)

**Describe the solution you'd like**
Add a separate config parameter for logging failed sql. Should be fairly simple.

**Describe alternatives you've considered**
It is possible to override the entityManager loader in containerServices.json and create a new EntityManagerFactory class that $logFailed to the DefaultSqlExecutor from a different config parameter  but it's a lot of code for such a little thing.

**Additional context**

https://github.com/espocrm/espocrm/blob/6b9ebdc731d09f48f7a6764108ad048d9505775f/application/Espo/ORM/Executor/DefaultSqlExecutor.php#L50-L51

https://github.com/espocrm/espocrm/blob/6b9ebdc731d09f48f7a6764108ad048d9505775f/application/Espo/Core/ORM/EntityManagerFactory.php#L124-L129

SuchAFuriousDeath:
@yurikuzn You might've missed this because I posted it on the weekend.

yurikuzn:
I haven't missed. The issue is there I'll look into when get a chance, soon or later. It's not a critical bug that is needed to be fixed asap.

PR #3133, opened 2024-08-29
Log only failed SQL queries

bandtank:
As described in #3110, here is a PR to log only failed SQL queries. I spent quite a while tracing the code and testing without finding any issues, but maybe I am not considering something. I am submitting the PR in case it is useful.

@SuchAFuriousDeath I think this is what you had in mind, but please let me know if anything is missing.


SuchAFuriousDeath:
Exactly what I had in mind

PR #3135, opened 2024-09-02
Fix template markup

tim-bezhashvyly:


yurikuzn:
I applied the fix to the *fix* branch. https://github.com/espocrm/espocrm/commit/7d507460be902ec387758a818dcb24ad86bb06af

yurikuzn:
BTW. How did you notice the problem?

tim-bezhashvyly:
The problem is not manifesting itself during rendering. I saw it in my IDE.

tim-bezhashvyly:
BTW I would appreciate if you will give me a hint how to override this template.

yurikuzn:
You can monkey patch the class from `views/site/header.js`.

PR #3138, opened 2024-09-08
admin ACL for clientDefs.json schema

Kharg:
add Admin ACL for clientDefs, I don't know if it's supported anywhere else apart from buttons and dropdowns, it would be nice to have it where it's supported so when using Schema it doesn't show the parameter as wrong.

### DIFF
--- a/schema/metadata/clientDefs.json
+++ b/schema/metadata/clientDefs.json
@@ -555,6 +555,7 @@
                                     "read",
                                     "edit",
                                     "delete",
+                                    "admin",
                                     "stream"
                                 ]
                             },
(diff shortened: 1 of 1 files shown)


yurikuzn:
There's no such an option.

yurikuzn:
Use *accessDataList* for admin check.

Kharg:
Thank you. It must have been a false memory.

Is that how the advanced pack does it for buttons? Using accessDataList when you select admin?

yurikuzn:
I don't remember. Maybe. It's also possible to use a handler to control visibility.

PR #3140, opened 2024-09-11
fix row action return undefined

eymen-elkum:


eymen-elkum:
Sorry for the mistake, fixed the base

PR #3157, opened 2024-09-19
code improvements

eymen-elkum:


yurikuzn:
I avoid direct usage of parameter values in class names. From my perspective, it's not an improvement. It would be an implicit support of custom parameter value which is not documented/officially supported. Maybe there's another way to achieve what you need.

arkadiyasuratov:
Just throwing my two cents, but I prefer the original version. Sure, it's somewhat verbose, but I find it more readable.

eymen-elkum:
I see, it is hard to get one vision on such topics :)

But was good to hear your comments for future's contributions.

yurikuzn:
Having explicit class names is also useful to be able to find it in codebase.

PR #3164, opened 2024-09-25
compose.js super.afterSave

Kharg:
This pull request fixes compose.js file where the afterSave() method was incorrectly calling super.afterRender() instead of super.afterSave().

When this was still an AMD module it was calling Dep.prototype.afterSave.call(this);

https://github.com/espocrm/espocrm/commit/d7063a914f9d0112b1905fe70245b493c0c99a4d#diff-117135453cb4900bd25d798364599ab10aa07e35722300e367a69b9df90e85a3L216

### DIFF
--- a/client/src/views/email/record/compose.js
+++ b/client/src/views/email/record/compose.js
@@ -215,7 +215,7 @@ class EmailComposeRecordView extends EditRecordView {
     }
 
     afterSave() {
-        super.afterRender();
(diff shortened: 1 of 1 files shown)


yurikuzn:
Might be better to apply to the fix branch. But it may bring some unexpected behavior. This mistake where there for more than a year.

PR #3207, opened 2024-11-13
Allow to customise the order of the activities panel

rabiibrahimi:
Currently the activities are ordered by dateStart and ordered in descending mode. 
However if someone wants to change that it won't work even if they use a custom activities view and provide an orderBy and order properties. This fix the issues and allow the user to have that flexibility so they could customise how the records are ordered on the activities panels.


yurikuzn:
I'll look later. I'm not quite sure that there's are no nuances that should be handled. For example, passing non-existent column not producing an appropriate exception.

rabiibrahimi:
> For example, passing non-existent column not producing an appropriate exception.

True if modifiedAt passed it produce an error. 

Leave it with you when you have time :)

Thanks

tovoro:
Any news here? Our Employees are requesting this, would be great to have this functionality!

PR #3219, opened 2024-11-24
Target list row actions translate

Kharg:
Row actions in Target Lists were not using translations.
Added this.translate to the labels.


Kharg:
This is how it appears right now in the demo:
image
image

Kharg:
Fixing that, thank you.

yurikuzn:
This is a proper way. Thank you.

PR #3236, opened 2024-12-07
[WIP] Fix italian translations

gremo:
Thank you for your work 🙏 . I am Marco, a Symfony programmer with 10 years of experience. I am getting closer to EspoCRM, taking my first steps.

I would like to correct the Italian translations. Besides typos and incorrect terms, many strings in Italian have incorrect capitalization. In the Italian language, it is rare to use capital letters; they are only used for proper names, languages, institutions, acronyms, and abbreviations, and very few other cases. Translations like "Token di Autenticazione" (authToken) are completely wrong.

It is a long task, but I am willing to complete it in a few days, if the PR gets accepted. 

This is just a WIP for now, awaiting your confirmation.


yurikuzn:
Hi Marco,

Thanks for your interest. Could you accept our CLA. Just a simple PR in the [CLA repo](https://github.com/espocrm/cla).

We use the [POEditor](https://github.com/espocrm/espocrm?tab=readme-ov-file#language) service for translations. Contributors add/fix translations there, and then we run a script that generates JSON files which we commit to this repo. Having this routine, there's a risk that some translation will be lost if we apply changes to JSON files first. Though it's possible. I would need to merge all the current translations from POEditor to here first. Then, during the period of making changes here, we should not change the Italian language in POEditor, otherwise, those changes can lost when importing from here to POEditor.

@Kharg Please let us know if you have some thoughts on this.

yurikuzn:
I've just applied the last changed from POEditor: https://github.com/espocrm/espocrm/commit/28715618bc97b5e506206ab3ec2399993d038a21.

Kharg:
Honestly I'd like to be ready to translate 9.0 from POEditor as soon as possible, and as far as I am understanding this will not be possible for a certain amount of time if we merge this PR?

I tried to remain faithful to the source material when translating into Italian, I've even asked some translators for their opinion about this capital initials issue and the general opinion was that since most of these are "titles" it was fine to use initials.
But I am open to change.

But please, do so in POEditor

gremo:
I can do it in POEditor (I think, never used it) if you'd like 👍 However, please let me know if I should start right now or wait for some decisions from you?

@Kharg about that: capitalization rules are generally stricter in Italian compared to English, and random use of uppercase within a sentence (except for proper nouns or specific cases) is considered incorrect.

yurikuzn:
In the original, we use upper case first letters for labels with few number of words. When there are many words, like a real sentence, we do use lower case first letters. Maybe the same rule is OK for Italian.

yurikuzn:
I've just checked Italian labels in another app. They do have Capital Case for short labels same as we do.  @Kharg How it's in Salesforce? :)

gremo:
@yurikuzn not always. An short label would be "Move to Trash" that is translated right now as  "Sposta nel Cestino" and this isn't correct. 

Here are a few exaples from Thunderbird (all lower as you can see):

- Download messages
- Open in a new windows
- Open in a new tab
- Search messages...

image

Gmail "Mark all as read":

image

Again thunderbird "Reply to the sender"/"Reply to all" (actually wrong in EspoCRM):

image

As I said, capita letters are very rare. Correct examples from Espo:

- "Create ApiUser": "Crea utente API"
- "PHP Settings": "Impostazioni PHP"
- "ldapCreateEspoUser": "Crea Utente in EspoCRM" ("Utente" is wrong, "EspoCRM" is a product name so it's fine)
-

Kharg:
The SaaS I work for is only capitalizing the first letter in its Italian translation.

I don't know about Salesforce as I use it in English :') 

I am open to adopt the right way of capitalizing.

yurikuzn:
Thanks @Kharg, I think we can change to the 'First word only capital'. Maybe it's reasonable to do it in automatic fashion somehow.

@gremo Please, we would need a CLA application to be able to accept language contributions.

gremo:
@yurikuzn not a good advice to do It automatically without a check by and human 😅

If you don't "trust" me, you can just wait for other italians working with It and confirming what I'm try to explain 😄 or you can ask me for screenshot or other example, no problem at all.

BTW out of curiosity, how the translations were created? Using some kind of automaric tool?

yurikuzn:
I'm OK with *Capital letter* change.

@Kharg I've uploaded the latest labels to POEditor. Please feel free to work with v9.0 translations if you'd like.

gremo:
Should we proceed? If yes, please authorize my POEditor account 😄 otherwise I'm gonna close this issue

PR #3246, opened 2025-01-19
Italian templates

Kharg:
Translations of the assignment/cancellation/invitation/reminder templates into Italian.


Kharg:
Hi Yuri,

I synced my fork before translating the template, so they are using dateStartFull as well.

Thank you for merging.

PR #3309, opened 2025-03-11
Add Select Appliers Support for Kanban View

eymen-elkum:
Hi Yuri,

This extends the Select Appliers feature to Kanban, ensuring it works like in normal record list. While Select Appliers is a great feature, it hasn't been implemented for Kanban yet—this update addresses that.

Looking forward to your review!

ref: https://github.com/espocrm/espocrm/pull/2629/files


yurikuzn:
Hi Eymen,

Yesterday I filed a backlog record about adding forced select appliers (in selectDefs rather than in recordDefs).

I believe it would solve the Kanban problem.

I'll need to think whether we need recordDefs appliers for Kanban, to make sure whether it can't introduce any issues.

yurikuzn:
We'd need to switch the base to the *master* branch instead of the *fix*.

PR #3318, opened 2025-03-19
Hide Plus button on kanban header when createDisabled is true

rabiibrahimi:
- Check if `createDisabled` is set to `true` in clientDefs.
- Hide the plus button in the kanban header when `createDisabled` is true.


yurikuzn:
It has to be in setup: https://github.com/espocrm/espocrm/commit/080a7864b5c5674d57615f5cd889d7f67bac98dd

PR #3327, opened 2025-03-29
Add Inject attribute for property injection

alexsvorada:
I've been thinking of this change as I don't really like the Aware/Setter pattern because it requires you to implement the pattern for your own services.

With this change it would be possible to do the following to inject a dependency that exists in the container:
```php
<?php

class MyService 
{
    public function doSomething(): string
    {
        return 'test';
    }
}
```

```php
<?php

use MyService;
use Espo\Core\Di\Inject;

class Foo()
{
    #[Inject]
    private MyService $myService;

    public function bar(): void;
    {
        $test = $this->myService->doSomething();
    }
}
```

### DIFF
--- a/application/Espo/Core/Di/Inject.php
+++ b/application/Espo/Core/Di/Inject.php
@@ -1,0 +1,23 @@
+<?php
+
+namespace Espo\Core\Di;
+
+use Attribute;
+
+/**
+ * Attribute for property injection.
+ * Can be used on properties to inject services from the container.
+ * 
+ * If a service name is not specified, it will be derived from the property type.
+ */
+#[Attribute(Attribute::TARGET_PROPERTY)]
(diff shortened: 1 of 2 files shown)


yurikuzn:
The Aware/Setter way is  for legacy code where inheritance is used. 

The approach you propose is not desirable from my perspective. The constructor injection should be used. It's quite of a core change I'd avoid.

yurikuzn:
Why not:

* Invalid class state after constructor.
* Can't pass dependencies when testing.
* An extra way that won't be used in our codebase.

alexsvorada:
I definitely agree with mostly using constructor injection, just in cases where property injection is needed this seems like a cleaner option.

Anyways, thank you for your answer!

PR #3328, opened 2025-03-31
Fix Regex value limited by field's maxLength

alexsvorada:
When a field has maxLength defined, it also affects regex length in dynamic/conditional logic.

This fix overrides the maxLength for regex to '100' and truncates the value when switching off of regex option


yurikuzn:
I refactored and fixed it: https://github.com/espocrm/espocrm/commit/e9fbb1d4d454b5dab1bbb3e9671faa722d08fea3

https://github.com/espocrm/espocrm/commit/e9fbb1d4d454b5dab1bbb3e9671faa722d08fea3

PR #3358, opened 2025-05-07
fix typo User.json

Sovicka:
fix typo


yurikuzn:
I fixed on POEditor. As we use it as a source of truth for translations.

PR #3369, opened 2025-05-14
fixes TIMESTAMPDIFF_MONTH error for postgresql

SuchAFuriousDeath:
Seems to be a clear typo


yurikuzn:
Thank you for the fix.

PR #3371, opened 2025-05-17
Fixes wrong translation scope

SuchAFuriousDeath:


SuchAFuriousDeath:
The other upload button is translated from the Admin scope:

https://github.com/espocrm/espocrm/blob/2aacc5277d1bcc3ca5f0cd09e2d2da20b23605a1/client/res/templates/admin/extensions/index.tpl#L19

SuchAFuriousDeath:
By the way, I requested to join the POEditor project for Czech. There is a lot missing and even what is translated seems quite awkward at times. I'm hoping to fix that.

yurikuzn:
Thank you. I accepted you in POEditor.

PR #3373, opened 2025-05-18
Update PL translations

mkz212:


yurikuzn:
Hi, Thanks for fixes.

We use the POEditor service for translations. All labels are translated there vie the UI. And then, merged to here. The POEditor translations are the source of truth. It's bit problematic to accept translations here and then upload to the POEditor.

We also require acceptance of CLA for contributions.

PR #3421, opened 2025-06-24
French e-mail template for password reset is now translated

tbrugier:
I translated the `.tpl` file containing the e-mail body that's sent to the user resetting its password.


yurikuzn:
Thanks.

PR #3440, opened 2025-07-24
Include isResolved in cancelCallback of Espo.Ui.confirm

alexsvorada:
This would a nice QoL feature so that we can detect & perform different logic in cancelCallback based on if the user canceled using the button or ESC/clicking outside the button.


yurikuzn:
Style changes are [not to be included](https://github.com/espocrm/espocrm/blob/master/.github/CONTRIBUTING.md#pull-requests) in PRs. 

I don't think we need a new 3rd state. Currently, we have Yes or No. No other option. I believe it's better to implement a custom confirmation for such a case.

alexsvorada:
My bad I didn't think it was a style change. As for "3rd" state, it can be fully ignored if not needed but I understand it's probably better to implement a custom confirmation

PR #3451, opened 2025-08-15
keep user-addressed emails visible when routed to a Group Folder

traien:
**Describe the bug**
When an incoming email is addressed to a user's personal email (To:) and the email is CC'd to a group email that has a Group Folder configured, the mail processing moves the email to that Group Folder. If the user does not have access to the Group Folder, the user cannot see the email in their inbox nor in the Group Folder.

**To Reproduce (MANDATORY, DO NOT REMOVE)**
1. Create/configure a Group email account (e.g. group@example.com) and assign it a Group Folder (e.g. "Sales").
2. Ensure a user account (e.g. user@example.com) does *not* have access permissions to that Group Folder.
3. Send an email where:
   - To: user@example.com
   - CC: group@example.com
4. Observe that the email is moved into the Group Folder and no copy appears in the user's personal inbox. Because the user lacks access to the Group Folder, they cannot locate the email anywhere in EspoCRM.

**Observed / Actual behavior**
The email is moved to the Group Folder (assigned to the CC'd group account) and therefore disappears from the user's inbox. Since the user lacks permissions on that Group Folder, they cannot view the email at all.

**Expected behavior**
If a recipient (To:) does not have access to a Group Folder that an email would be routed into (because of a CC'd group account), EspoCRM should **not** silently filter/move the email out of the recipient's inbox. Possible acceptable behavior:
Show the email to both the recipient's inbox and the Group Folder

**EspoCRM version**
Please specify the EspoCRM version(s) where this occurs (e.g. `9.1.8`).

**Additional context**

This change updates the inbox filter to include emails that are addressed to the current user even when groupFolderId is present.

### DIFF
--- a/application/Espo/Classes/Select/Email/Where/ItemConverters/InFolder.php
+++ b/application/Espo/Classes/Select/Email/Where/ItemConverters/InFolder.php
@@ -30,10 +30,12 @@
 namespace Espo\Classes\Select\Email\Where\ItemConverters;
 
 use Espo\Core\Name\Link;
+use Espo\Core\Select\SelectBuilderFactory;
 use Espo\Core\Select\Where\ItemConverter;
 use Espo\Core\Select\Where\Item;
 
 use Espo\Entities\Email;
+use Espo\Entities\GroupEmailFolder;
 use Espo\ORM\Name\Attribute;
 use Espo\ORM\Query\SelectBuilder as QueryBuilder;
 use Espo\ORM\Query\Part\WhereItem as WhereClauseItem;
@@ -51,7 +53,8 @@ class InFolder implements ItemConverter
     public function __construct(
         private User $user,
         private EntityManager $entityManager,
-        private JoinHelper $joinHelper
+        private JoinHelper $joinHelper,
+        private SelectBuilderFactory $selectBuilderFactory
     ) {}
 
     public function convert(QueryBuilder $queryBuilder, Item $item): WhereClauseItem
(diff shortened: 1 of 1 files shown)


yurikuzn:
The Inbox should not display Group emails, it was intentional. It breaks the designed behavior for users who have access to the group folder.

yurikuzn:
The proper fix would be to obtain the list of all group folders the user does not have access to and include them in the query. To display all where group ID is null or among not accessible.

yurikuzn:
Joining tables is very undesirable here. It could impact on performance of an already heavy query.

traien:
> The proper fix would be to obtain the list of all group folders the user does not have access to and include them in the query. To display all where group ID is null or among not accessible.

I tried to achieve that in my last commit

yurikuzn:
Need to set user, as the selectBuilderFactory is registered as service, an can be run in a context where the current user is the system user or other user.

```
$selectBuilder->forUser($this->user)
```

Also we always use `private` rather than `protected`.

Thanks.

yurikuzn:
Thanks. Looks fine. Hope it won't affect performance much.

yurikuzn:
The fix was not functional. The select access control filter was not implemented for the GroupEmailFolder. For regular users, it displayed all emails. regardless of not having access to group folders.

Fix: https://github.com/espocrm/espocrm/commit/b78e87b3e3d7474a5cb1ea515ec4853b3d64559b

PR #3469, opened 2025-09-09
Fix Htmlizer translation for enum fields

AntonSlyzhko:
Using original value of enum option instead of attribute name. Also tested with null.


yurikuzn:
Nice. Will need also checking whether the raw value is not null.

PR #3472, opened 2025-09-18
fix applying 'inCategory' filter

AntonSlyzhko:
Previous condition was never satisfied, so 'inCategory' was never applied


yurikuzn:
I think we better remove the dead code. It will keep the functionality as it's now. The side panel can be used for In Category filtering, the field filter can be used for the exact match search.

https://github.com/espocrm/espocrm/commit/9aa3e3213f1687c36fd429ea8903c51439c5358b

PR #3473, opened 2025-09-24
load additional fields for parent entity in email template processor

traien:
This change ensures that the fields of the parent record are properly loaded.
For example, consider a template like this:

```
You are welcome to our Portal!

Parent: {Task.email.parentName}
```

Previously, if you created an email with a task as its parent — and that task also had an email with its own parent — this placeholder would return an empty value.

With this fix, the value now loads correctly.


traien:
Hi @yurikuzn ,

Can you review this PR?

yurikuzn:
I've been very busy for last month+ and likely will still be for a while.

This change in not a non-brainer. Nested links was not loaded for a reason.

PR #3475, opened 2025-09-25
fix clientDefs schema 

arkadiyasuratov:
I changed the `buttonList` type from an object to an array, because it should match the type of `actionList`.

https://github.com/espocrm/espocrm/blob/9.1.9/client/src/views/record/panels-container.js#L53-L54


yurikuzn:
Proper fix: https://github.com/espocrm/espocrm/commit/74cc70145974b4f74ec92918ded18e77172ec0e7

PR #3476, opened 2025-09-25
int field view - fix typo

AntonSlyzhko:
max is undefined in this scope, so condition is never satisfied


yurikuzn:
It does not have effect since we incorporated the library usage. https://github.com/espocrm/espocrm/commit/2465a554f7032c940563ceea62846e582d4d9617

PR #3482, opened 2025-10-01
statusField: Only storable enum

rabiibrahimi:
This pr is to allow statusField to be filled with only storable enum fields of the entitytype. 

Currently if you go to contact and edit the entity and choose a statsuField (Opportunity Role || Acceptance Status (Call) || Acceptance Status (Meeting)) end then enable kanban then kanban would appear first time but as soon as you change something in contact and try to view kanban - it is empty. The reason is that Opportunity Role is not storable field. 

I hope this is clear. 

Thanks

### DIFF
--- a/client/src/views/admin/entity-manager/edit.js
+++ b/client/src/views/admin/entity-manager/edit.js
@@ -184,7 +184,9 @@ class EntityManagerEditView extends View {
                         return;
                     }
 
-                    if (fieldDefs[item].type === 'enum') {
+                    if (fieldDefs[item].type === 'enum' &&
+                        (!('notSortable' in fieldDefs[item]) || fieldDefs[item].notSortable === false)
+                    ) {
                         return true;
                     }
                 })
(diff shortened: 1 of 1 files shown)


client/src/views/admin/entity-manager/edit.js
```diff
@@ -184,7 +184,9 @@ class EntityManagerEditView extends View {
                         return;
                     }
 
-                    if (fieldDefs[item].type === 'enum') {
+                    if (fieldDefs[item].type === 'enum' &&
+                        (!('notSortable' in fieldDefs[item]) || fieldDefs[item].notSortable === false)
```
SuchAFuriousDeath, line 188: I believe this should be 'notStorable'

rabiibrahimi: my bad. autocompletion - fixed now
Thanks

yurikuzn:
Hi. It's reasonable fix. I applied it to the fix branch, slightly diffently. I prefer to keep the 'notStorable' param away from the frontend. Thank you. https://github.com/espocrm/espocrm/commit/9153fbf280e183fdb68731d7145ebf0851dcd8d7

rabiibrahimi:
Thanks for the fix 👍

PR #3486, opened 2025-10-04
Allow system templates to be customized using metadata

bandtank:
This change allows developers to store custom templates in a module. For example:

`custom/Espo/Modules/MyCustomModule/Resources/metadata/app/templates.json`:
```
{
    "passwordChangeLink": {
        "scope": "User",
        "module": "MyCustomModule"
    }
}
```

### DIFF
--- a/application/Espo/Core/Utils/TemplateFileManager.php
+++ b/application/Espo/Core/Utils/TemplateFileManager.php
@@ -32,13 +32,15 @@ namespace Espo\Core\Utils;
 use Espo\Core\Utils\File\Manager as FileManager;
 use Espo\Core\Utils\Resource\FileReader;
 use Espo\Core\Utils\Resource\FileReader\Params as FileReaderParams;
+use Espo\Core\Utils\Metadata;
 
 class TemplateFileManager
 {
     public function __construct(
         private Config $config,
         private FileManager $fileManager,
-        private FileReader $fileReader
+        private FileReader $fileReader,
+        private Metadata $metadata,
     ) {}
 
     public function getTemplate(
@@ -48,9 +50,19 @@ class TemplateFileManager
         ?string $defaultModuleName = null
     ): string {
 
+        $templates = $this->metadata->get(['app', 'templates']);
+
+        $moduleName = null;
+
+        if (isset($templates[$type]) && isset($templates[$type]["module"])) {
+            $moduleName = $templates[$type]["module"];
+        } else {
+            $moduleName = $defaultModuleName;
+        }
+
         $params = FileReaderParams::create()
             ->withScope($entityType)
-            ->withModuleName($defaultModuleName);
+            ->withModuleName($moduleName);
 
         if ($entityType) {
             $path1 = $this->getPath($type, $name, $entityType);
--- a/application/Espo/Core/Utils/TemplateFileManager.php
+++ b/application/Espo/Core/Utils/TemplateFileManager.php
@@ -52,12 +52,10 @@ class TemplateFileManager
 
         $templates = $this->metadata->get(['app', 'templates']);
 
-        $moduleName = null;
+        $moduleName = $defaultModuleName;
 
         if (isset($templates[$type]) && isset($templates[$type]["module"])) {
             $moduleName = $templates[$type]["module"];
-        } else {
-            $moduleName = $defaultModuleName;
         }
 
         $params = FileReaderParams::create()


rabiibrahimi:
this is already possible for developers 

https://docs.espocrm.com/development/metadata/app-templates/

bandtank:
That doesn't work because the template file manager ignores the module parameter.

rabiibrahimi:
you meant to allow developers to customise existing template i guess.

bandtank:
That is correct. For example:
`application/Espo/Resources/metadata/app/templates.json`:
```
{
    "accessInfo": {
        "scope": "User"
    },
    "accessInfoPortal": {
        "scope": "User"
    },
    "assignment": {
        "scopeListConfigParam": "assignmentEmailNotificationsEntityList"
    },
    "mention": {
        "scope": "Note"
    },
    "noteEmailReceived": {
        "scope": "Note"
    },
    "notePost": {
        "scope": "Note"
    },
    "notePostNoParent": {
        "scope": "Note"
    },
    "noteStatus": {
        "scope": "Note"
    },
    "passwordChangeLink": {
        "scope": "User"
    },
    "twoFactorCode": {
        "scope": "User"
    }
}
```

It is impossible to override the templates shown above without the changes in this PR.

rabiibrahimi:
I totally understand however since the end user can change the template from the UI why would a developer need this?

bandtank:
The changes to those files are not saved in the database. Customizing the files while using containers means needing to copy the files to every instance in real time. Putting the files in an extension is how I currently do it, and it works very well.

yurikuzn:
I'll look into more detail when get a chance. I think we might remove *$defaultModuleName* too, but it would be a small breaking change.

bandtank:
As far as I can tell, the commit I pushed a moment ago removes the parameter to no effect.

My assumption is you want to keep the `module` parameter in `templates.json` because it is needed by PHP to find the template files. I do not think it is necessary to pass the `module` parameter to the template manager because the value can be found in metadata using the changes in this PR. To that end:

These files use the `TemplateFileManager` class, but not the `defaultModuleName` parameter in `getTemplate`:
```
application/Espo/Core/Di/TemplateFileManagerAware.php
application/Espo/Core/Di/TemplateFileManagerSetter.php
application/Espo/Core/Authentication/TwoFactor/Email/Util.php
application/Espo/Tools/EmailNotification/Processor.php
application/Espo/Tools/EmailNotification/AssignmentProcessor.php
application/Espo/Tools/UserSecurity/Password/Sender.php
application/Espo/Tools/UserSecurity/Password/RecoveryService.php
application/Espo/Resources/metadata/app/containerServices.json
```
These files use the class and the `defaultModuleName` parameter:
```
application/Espo/Core/Utils/TemplateFileManager.php
application/Espo/Controllers/TemplateManager.php
application/Espo/Modules/Crm/Tools/Reminder/Sender/EmailReminder.php
application/Espo/Modules/Crm/Business/Event/Invitations.php
```
I removed the parameter from the call in the latter three files and then put a debug statement in `TemplateFileManager->getTemplate()`:
```
$this->log->debug("$type, $name, $entityType, module name: $moduleName", [$defaultModuleName]);
```

Everything worked as expected, and here are the log statements:
```
# Sending invitations
#   Original
[2025-10-06 14:22:50.539211] DEBUG: invitation, subject, Call, module name: Crm ["Crm"]
[2025-10-06 14:22:50.539994] DEBUG: invitation, body, Call, module name: Crm ["Crm"]
#   Parameter removed
[2025-10-06 14:24:07.569370] DEBUG: invitation, subject, Call, module name: Crm [null]
[2025-10-06 14:24:07.569990] DEBUG: invitation, body, Call, module name: Crm [null]

# Sending reminders
#   Original
[2025-10-06 14:32:05.226589] DEBUG: reminder, subject, Call, module name: Crm ["Crm"]
[2025-10-06 14:32:05.227502] DEBUG: reminder, body, Call, module name: Crm ["Crm"]
#   Parameter removed
[2025-10-06 14:40:06.810064] DEBUG: reminder, subject, Call, module name: Crm [null]
[2025-10-06 14:40:06.810665] DEBUG: reminder, body, Call, module name: Crm [null]

# Loading templates in the Template Manager
#   Original
[2025-10-06 14:45:22.580132] DEBUG: twoFactorCode, body, , module name:  [null]
[2025-10-06 14:45:22.582695] DEBUG: twoFactorCode, subject, , module name:  [null]
[2025-10-06 14:45:25.553817] DEBUG: accessInfo, body, , module name: Odyssey ["Odyssey"]
[2025-10-06 14:45:25.554197] DEBUG: accessInfo, subject, , module name: Odyssey ["Odyssey"]

#   Parameter removed
[2025-10-06 14:45:53.709840] DEBUG: twoFactorCode, body, , module name:  [null]
[2025-10-06 14:45:53.710399] DEBUG: twoFactorCode, subject, , module name:  [null]
[2025-10-06 14:45:55.482377] DEBUG: accessInfo, body, , module name: Odyssey [null]
[2025-10-06 14:45:55.482775] DEBUG: accessInfo, subject, , module name: Odyssey [null]
```
---

The question remains if you want to save changes to templates in `custom/Espo/Custom` instead of the original location, such as in an extension. This happens in `TemplateFileManager->saveTemplate()`. I think the behavior should not change because the templates in an extension should be treated as read-only. Perhaps disallowing edits to templates in extensions would be the right thing to do.

yurikuzn:
My concern was only about extensions. I believe it does not break our recent extension versions as we have the module set in metadata for all templates we use (though needs a thourough revision to be sure). But it would break 3rd party extensions (with quite a high chance I'd estimate).

> templates in an extension should be treated as read-only

All our templates defined in extensions are editable.

bandtank:
I see what you mean. My extension uses this feature and there are no issues, but I don't call the template file manager in PHP. My templates.json file uses the module parameter, and it works well. Without the module parameter, it works only if the file exists in custom/Espo/Custom, which is obviously a problem in many cases.

Re: the read-only edits - Ok, that's not a problem and it makes sense to me. My intent was to offer a solution to a potential problem (editing a template from an extension and then having a new file be saved in custom/Espo/Custom, which would need to be copied in real time to other instances of Espo), but if it's not a problem, then nevermind.

bandtank:
The update has been pushed. Sorry about that. I manually tested the changes as much as I could.

yurikuzn:
Fine, Thanks for the PR.

We might also need to inform about small breaking change in the next release in release notes.

Issue #3493, opened 2025-10-14
Uninstalling extension fails when client/custom and custom are symlinks


BrutalBirdie:
**Describe the bug**
Running:
```bash
php command.php extension -u --name="Advanced Pack"
```
returns:
```bash
Uninstalling... Do not close the terminal. This may take a while...
Permission denied to delete files.
```

**To Reproduce**

I am running EspoCRM on Cloudron, but the reproduction steps should work either way.

Explicit steps to reproduce the behavior:
1. make `client/custom` and `custom` a symlink
2. install the extension `Advanced Pack`
3. try to uninstall the extension with `php command.php extension -u --name="Advanced Pack"`

**Expected behavior**
symlinks should be handled accordingly.

**Screenshots**

**EspoCRM version**
Espo CRM 9.2.2

**Additional context**
Running Espo CRM 9.2.2 on Cloudron.
A PR is on its way to handle symlinks in `rmdir()`

yurikuzn:
Hi. I think we should treat it as a feature request, not a bug.

bandtank:
@BrutalBirdie What is the use case for using symlinks? I am wondering if storing the files elsewhere is advantageous for some reason.

BrutalBirdie:
Hello @bandtank and sorry for the delayed response.

The context of Cloudron I could have explained better.

Cloudron runs apps like espocrm (docker images) in [ isolation and sandboxing](https://docs.cloudron.io/security/#app-isolation-and-sandboxing).
Apps run with a read-only rootfs preventing attacks where the application code can be tampered with.
So all the user-data that needs to be persistent has to be placed in `/app/data` while the application is in `/app/code`.
This also reduces the backup footprint to what is really needed instead of the whole application with every backup.

If apps can't configure storage locations for user-data we have to work with symlinks to move the data to `/app/data`.

For espocrm in the Cloudron [Dockerfile](https://git.cloudron.io/packages/espocrm-app/-/blob/master/Dockerfile?ref_type=heads) we have to do some symlinks like:

```Dockerfile
RUN rm -rf /app/code/current/data && ln -s /app/data/data /app/code/current/data && \
    rm -rf /app/code/current/custom && ln -s /app/data/custom /app/code/current/custom && \
    rm -rf /app/code/current/client/custom && ln -s /app/data/client/custom /app/code/current/client/custom && \
[...]

PR #3494, opened 2025-10-14
Only run rmdir() when not is_link()

BrutalBirdie:
This is a suggested fix for #3493

### DIFF
--- a/application/Espo/Core/Utils/File/Manager.php
+++ b/application/Espo/Core/Utils/File/Manager.php
@@ -701,7 +701,7 @@ class Manager
         $result = true;
 
         foreach ($dirPaths as $dirPath) {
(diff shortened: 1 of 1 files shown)


yurikuzn:
Hi,

I'm not sure that it's how we should fix it. It skips removal when the passed path is not a directory. I assume that we should delete the directory by its real path. But it needs more thorough check. As it's the very core function.

BrutalBirdie:
Hello @yurikuzn 
True.
I made this quick and dirty fix for me to resolve two tickets we got in the Cloudron Forum.
Also, I never really did much PHP development, so I am not the best choice to provide a solid solution 😅

But I can get the stone rolling 😉

PR #3498, opened 2025-10-21
Fix OAuth refresh token

traien:
The logic of checking refresh token expiration is inverted


yurikuzn:
Thank you. I ported the fix to the *fix* branch.

Issue #3505, opened 2025-10-30
Espo\Tools\DataPrivacy\Erasor tries to access protected property `id` on entities


bfzgaier:
**Describe the mistake**
When handling the erason of emailAddresses and phoneNumbers the erasor directly access the entity id.
Because the property is not public, an exception is thought. 
Here is an example for `EmailAddress`:
`ERROR: Slim Application Error Type: Error Code: 0 Message: Cannot access protected property Espo\Entities\EmailAddress::$id File: /espo/application/Espo/Tools/DataPrivacy/Erasor.php Line: 117`

**Location**
https://github.com/espocrm/espocrm/blob/master/application/Espo/Tools/DataPrivacy/Erasor.php#L117

https://github.com/espocrm/espocrm/blob/master/application/Espo/Tools/DataPrivacy/Erasor.php#L135

**Suggested correction**
L117:  `$emailAddress->set(Field::NAME, 'ERASED:' . $emailAddress->getId());`
L135:  `$phoneNumber->set(Field::NAME, 'ERASED:' . $phoneNumber->getId());`

**Additional context**
none

PR #3506, opened 2025-10-30
Fix issue 3505 - Espo\Tools\DataPrivacy\Erasor tries to access protected property id on entities

bfzgaier:
This fixes the issue https://github.com/espocrm/espocrm/issues/3505

Sorry, for maybe selecting the wrong branch.


yurikuzn:
Thank you for the fix. I'll port it to the *fix* branch.

PR #3507, opened 2025-10-30
int getMinValue fix

antonslyzhko:
Fix getMinValue logic that overwrote valid minValue due to wrong null comparison


yurikuzn:
I think we'll need to simplify this and other such usages to only params, as it's supposed to be populated automatically from the  model parameters.

PR #3520, opened 2025-11-14
User stream: Fix incorrect alias in subscriptionExclude join

antonslyzhko:


yurikuzn:
Could you provide more information? It's important functionally affected. Did you test the result?

yurikuzn:
Could you also tell whether you observed a wrong behavior. What exactly wrong behavior.

antonslyzhko:
I'm currently researching how to optimize user stream loading. We also encountered a bug related to slow stream load, but that issue is not part of this PR.

While analyzing the SQL generated by the stream query to determine which indexes could be added, I noticed this `LEFT JOIN`. The condition `subscription.user_id = 'someUserId'` inside the `subscriptionExclude` join cannot have any effect, because it is already guaranteed to be true by the previous `INNER JOIN subscription`, which uses the same `user_id` condition. As a result, the `user_id` filter is applied to the wrong alias and does not actually restrict `subscriptionExclude` to the current user.

This PR corrects that by applying `someUserId` to the proper join alias, so the exclusion logic works as intended.

yurikuzn:
I see that the query is not correct. But was curious about actual reproducing. Whether the issue is observable. If yes, then what exactly is displayed wrong or not displayed when should be displayed. Whether you could observe the result of the fix.

We don't have problems with stream performance btw. Performance problems are issues, not bugs. It can be that some query performs much slower on a particular database and fast on another.

yurikuzn:
I believe that the wrong result would be a entry duplicates in the user's stream. The user needs to be subscribed to both Account and the related record, and multiple users have to be subscribed to the same record in order this bug to be observable

PR #3531, opened 2025-11-27
fix salutation name hardcoded

antonslyzhko:


yurikuzn:
As it's not a bug nor a wrong behavior, it's rather be in the dev branch.

yurikuzn:
To be able to merge, need rebase against master.

yurikuzn:
I'd define a protected property in the view class. Having the property defined make it less prone to changes in future.


```
/**
 * @protected
 * @type {string}
 * @since 9.3.0
 */
salutationField
```

yurikuzn:
I added to the master. https://github.com/espocrm/espocrm/commit/d179b4b792d0e24c4930eab03957dec86cba2a2d

PR #3545, opened 2025-12-14
Added clientReadOnly to schema

Daliborlol:


yurikuzn:
This was not added on-purpose, as being a non-official parameter. Maybe needs to be re-considered. Not yet sure,

PR #3555, opened 2026-01-03
fix required field validation

antonslyzhko:
This PR fixes incorrect validation logic for fields whose required flag is changed after records already exist.

If a field is made not required, records are created, and then the field is made required again, EspoCRM allows updating other fields via inline edit while bypassing validation of the newly required field.
This results in records being saved in an invalid state.

**Steps to reproduce**
1. In **Entity Manager**, make a field **not required**.
2. Create a record of this entity type.
3. Change the same field back to **required** in **Entity Manager**.
4. Open the previously created record.
5. Edit any _other_ field using **inline edit** (do not touch the required field).
6. Save — the update succeeds even though the required field is empty.


yurikuzn:
It's an expected behavior. Only updated fields are checked in field validation unless forced with a parameter. The if-block addresses this logic.

PR #3561, opened 2026-01-15
Update PreferencesService to consider dashboard layout metadata

kenyonit:
The resetDashboard function only considers the dashboardLayout and dashletsOptions config. 

Updated to check metadata app.defaultDashboardLayouts.Standard and app.defaultDashboardOptions.Standard so behaviour is consistent with the loadData function in Preferences.php


yurikuzn:
Hi. Could you accept the [CLA](https://github.com/espocrm/cla) ?

kenyonit:
Hi @yurikuzn, done mate

PR #3567, opened 2026-01-23
prevent NotImplemented exception during template rendering

antonslyzhko:


yurikuzn:
PRs are expected to have bit more input.

Not specifying the action is not good, one cannot rely on internals. This fix potentially lifts the permissions by not specifying the action.

antonslyzhko:
We experience this error with entity type which has only entityDefs metadata defined, so this is a utility table

yurikuzn:
It happens even if there's no any placeholder for that entity?

I see it might happen when there's no scope defined in the metadata. I didn't observe this problem as usually I define scope. There are other potential places where the similar error will occur I think. I'd recommend to define a scopes/{Entity}.json with "entity": true.

antonslyzhko:
That entity is never mentioned in the templates. And yes, there is no scope defined.

antonslyzhko:
Got it, thanks. I will define scope then

PR #3573, opened 2026-02-06
Fix undefined stdClass property warnings when viewing notifications

antonslyzhko:
Fixes E_WARNING "Undefined property: stdClass::$relatedName / $createdByName"
triggered when opening notifications via the bell icon.

Notification::getData() returns stdClass and may not contain these
properties. Access is now guarded with isset() to prevent warnings.

### DIFF
--- a/application/Espo/Tools/Notification/RecordService.php
+++ b/application/Espo/Tools/Notification/RecordService.php
@@ -472,12 +472,14 @@ class RecordService
 
     private function prepareSetFields(Notification $entity): void
     {
-        if ($entity->getRelated() && $entity->getData()?->relatedName) {
-            $entity->set('relatedName', $entity->getData()->relatedName);
+        $data = $entity->getData();
+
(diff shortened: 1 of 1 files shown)


yurikuzn:
Thank you for the fix. I applied the same with bit different style: https://github.com/espocrm/espocrm/commit/2ef02e8144ceb911234cccf1de4ecbe91d2cf005

antonslyzhko:
`getData` method can return **null** and potentially break the app based on the environment. I would prefer to use **null-safe operator** after calling `getData`

```
$relatedName = $entity->getData()?->relatedName ?? null;
$createdByName = $entity->getData()?->createdByName ?? null;
```

yurikuzn:
The null-coalesce operator in PHP is designed for this. Phpstan produces an error for `$entity->getData()?->relatedName ?? null`, saying ? is not needed.

PR #3575, opened 2026-02-09
fix language

ShaikhNaasir:


yurikuzn:
Hi,

We use POEditor as a source of truth. Did you change the labels there as well? If yes, then I can apply the commit.

ShaikhNaasir:
Hi Yuri, in my POEditor there is an issue. Could you please do this commit now? Otherwise we have to wait 5-7 days.

yurikuzn:
There's no need to commit now, as there may be no release that soon.

Another problem, it's against the master branch. Should be the fix branch to be included in the hotfix release.

PR #3621, opened 2026-03-31
Fix: scheduled email stream note missing user attribution

### LINKED ISSUE #3541: Stream note for scheduled emails missing user attribution (personEntityId null)
**Describe the bug**
When an email is sent via the "Send Later" (scheduled) feature, the stream note on the related entity (Lead/Contact/etc.) shows "sent email" without the user name.

**To Reproduce (MANDATORY, DO NOT REMOVE)**
1. Create a Lead with an email address
2. Go to compose email from that Lead
3. Use "Send Later" to schedule the email for a future time
4. Wait for the scheduled email job to send it
5. View the Lead's stream panel - the note shows "sent email" without user name

**Expected behavior**
Stream note should show: "[User Name] sent email" with a clickable link to the user who scheduled the email.

**Screenshots**
N/A

**EspoCRM version**
9.2.5

**Additional context**
**Technical Analysis:**
In `application/Espo/Tools/Stream/Service.php`, the `noteEmailSent()` method (lines 552-570) uses `$this->user` to determine the person for attribution:

```php
$user = $this->user;  // System user when running via scheduled job
$person = null;

if (!$user->isSystem()) {
    $person = $user;  // Works for manual sends
} else {
    // For scheduled: tries to find person by email FROM address
    $from = $email->getFromAddress();
    if ($from) {
        $person = $this->getEmailAddressRepository()->getEntityByAddress($from);
        // Returns NULL if FROM address isn't linked to Contact/Lead/User
    }
}
```

When scheduled emails are sent via the `SendScheduledEmails` job, the service runs in system user context. The fallback logic tries to find a person by the email's FROM address, but this fails when using group email accounts (InboundEmail) where the FROM address isn't linked to a Contact/Lead/User entity.

**Suggested Fix:**
When in system context, use `$email->getCreatedBy()` to get the user who scheduled the email, before falling back to email address lookup:

```php
if (!$user->isSystem()) {
    $person = $user;
} else {
    // First try to get the user who created/scheduled the email
    $createdBy = $email->getCreatedBy();
    if ($createdBy) {
        $person = $this->entityManager->getEntityById(User::ENTITY_TYPE, $createdBy->getId());
    }

    // Fallback to email address lookup
    if (!$person) {
        $from = $email->getFromAddress();
        if ($from) {
            $person = $this->getEmailAddressRepository()->getEntityByAddress($from);
        }
    }
}
```

uszekkk:
## Summary

Fixes #3541

When emails are sent via the **Send Later** (scheduled) feature, the stream note on the related entity shows "sent email" without the user name. This happens because the `SendScheduledEmails` job runs in system user context, and the fallback logic that tries to find a person by the email's FROM address fails for group email accounts.

## Changes

In `application/Espo/Tools/Stream/Service.php`, method `noteEmailSent()`:

- Added a `getCreatedBy()` lookup as the **first fallback** when running in system context
- This resolves the user who originally composed/scheduled the email
- The existing FROM address lookup is preserved as a second fallback

## Before
```
"sent email" (no user name)
```

## After
```
"John Smith sent email" (with clickable link to user)
```

## Test plan
1. Create a Lead/Contact with an email address
2. Compose an email from that entity using "Send Later"
3. Wait for the scheduled job to send it
4. Verify the stream note shows the correct user name

### DIFF
--- a/application/Espo/Tools/Stream/Service.php
+++ b/application/Espo/Tools/Stream/Service.php
@@ -558,10 +558,18 @@ class Service
         if (!$user->isSystem()) {
             $person = $user;
         } else {
-            $from = $email->getFromAddress();
+            $createdBy = $email->getCreatedBy();
 
-            if ($from) {
-                $person = $this->getEmailAddressRepository()->getEntityByAddress($from);
+            if ($createdBy) {
+                $person = $this->entityManager->getEntityById(User::ENTITY_TYPE, $createdBy->getId());
+            }
+
+            if (!$person) {
+                $from = $email->getFromAddress();
+
+                if ($from) {
+                    $person = $this->getEmailAddressRepository()->getEntityByAddress($from);
+                }
             }
         }
 


yurikuzn:
Thanks for the proposed fix. Could you accept our CLA by making a PR in the [CLA](https://github.com/espocrm/cla) repository?

uszekkk:
CLA signed: espocrm/cla#120

PR #3639, opened 2026-04-13
replace this.name with this.link

antonslyzhko:
`this.name` is not initialized in `RelationshipPanelView`, so these references resolve to `undefined`.


yurikuzn:
Hi. Thanks for the PR. Could you change the branch to the *master*.

PR #3650, opened 2026-04-21
Fix backdrop option reference in confirm dialog

Kharg:


yurikuzn:
Actually, it should not be like this. It should check against 'o'  rather than read options.

Now, `o` is highlighted as unused which indicates that something likely wrong.

Kharg:
Your fix works. Thanks.

Sorry for my wrong fix, building from master and testing again and again takes a bit :)

yurikuzn:
Master is very unstable now. It's likely has bugs.

PR #3657, opened 2026-04-29
LDAP: add local auth fallback for regular users

### LINKED ISSUE #1948: LDAP Authentication fallback method
Briefly, It would be a nice feature to have if LDAP authentication method has a fallback when the user is not present in the LDAP tree.

As explained more here -> https://forum.espocrm.com/forum/general/62683-fallback-authentication-mechanism-when-ldap-fails
Basically this would allow us to use LDAP Auth + Espo Auth all together. 

In my case I also have a lot of contractors and partners who are not present in my AD LDAP.

kboykov:
Closes #1948.

Currently, when LDAP authentication is enabled, only admin users can fall back to local Espo authentication (when the LDAP server is unreachable or the user is not found in LDAP). Regular users — such as contractors and partners who are not present in the LDAP/AD tree — have no fallback at all. There is also no fallback of any kind when LDAP accepts a user's DN but rejects the password bind.

This PR adds an opt-in `ldapFallbackToLocalAuth` config flag (default: `false`, so existing behaviour is fully preserved). When enabled:

- Any non-API, non-system user with a local Espo password can authenticate via local auth if LDAP is unavailable.
- The same fallback applies when the user is not found in LDAP (the primary use case from #1948 — users such as contractors not present in AD).
- A new fallback covers the LDAP password-bind failure case, which previously had no fallback even for admins.

The existing `adminLogin()` method is left untouched. A new `localLogin()` method handles the broader user scope, with the same type exclusions as the downstream user lookup (`TYPE_API`, `TYPE_SYSTEM`, `TYPE_SUPER_ADMIN`).

### DIFF
--- a/application/Espo/Core/Authentication/Ldap/LdapLogin.php
+++ b/application/Espo/Core/Authentication/Ldap/LdapLogin.php
@@ -154,6 +154,8 @@ class LdapLogin implements Login
 
         $ldapClient = $this->getLdapClient();
 
+        $fallbackToLocal = (bool) $this->config->get('ldapFallbackToLocalAuth', false);
+
         /* Login LDAP system user (ldapUsername, ldapPassword) */
         try {
             $ldapClient->bind();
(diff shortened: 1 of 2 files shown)


kboykov:
Ref: https://forum.espocrm.com/forum/general/82292-local-users-beside-ldap
Ref: https://forum.espocrm.com/forum/general/62683-fallback-authentication-mechanism-when-ldap-fails
Ref Internal ticket: 1740

yurikuzn:
Hi,

Thanks for the PR.

I'm not sure it's good to replicate the same fallback mechanism that we have for admins for regular users. The existing mechanism that it falls back on authentication failure is not a nice approach. I'm not sure we need to continue with the same approach.

diusazzad:
Hi @kboykov, solid work on the LDAP fallback logic.Quick question regarding naming: In localLogin, we're using $adminUser even for non-admin users. It feels a bit misleading. Would you mind renaming that to just $user or $authenticatedUser to keep things clear?Otherwise, everything else (especially the #[SensitiveParameter] usage) looks great.

PR #3658, opened 2026-04-29
Add notification sounds for new notifications

kboykov:
Enables notification sounds for new in-app notifications.

The repository has partial sound-related scaffolding, but the new-notification badge path hard-codes sounds as disabled, so playback never happens. This PR exposes the setting in admin config and reads it at playback time.

### DIFF
--- a/application/Espo/Resources/defaults/config.php
+++ b/application/Espo/Resources/defaults/config.php
@@ -268,6 +268,7 @@ return [
     'auth2FAInPortal' => false,
     'personNameFormat' => 'firstLast',
     'newNotificationCountInTitle' => false,
+    'notificationSoundsDisabled' => true,
     'pdfEngine' => 'Dompdf',
     'smsProvider' => null,
     'mapProvider' => 'Google',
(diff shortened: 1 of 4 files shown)


kboykov:
Tested with Edge, Chrome, Firefox - notification sounds work for me. Can't test with Safari.

Previous implementation didn't work in any case

yurikuzn:
It actually adds notification sound for new notifications. We currently don't have notification sounds.

kboykov:
Yes, you are right, but it seems that this feature has been requested by a lot of people, dating back to 2020. It also seems that it is a hidden feature and currently it is hardcoded to be permanently disabled for reasons unknown to me

yurikuzn:
To implement this feature back, we will need to do:

1. Find a nice open sourced sound effect.
2. Implement settings parameter (enabled/disabled, not disabled = true).
3. Implement override in preferences.

Please, for next times, describe what PR does. If it enables a feature, it should state clearly that it enables the feature rather than fixing some problem but implicitly adds a feature.

yurikuzn:
https://github.com/espocrm/espocrm/issues/3661

PR #3687, opened 2026-06-02
Fix OIDC sync clearing user fields when claims are absent from the token

### LINKED ISSUE #3638: OIDC sync clears phone number when provider doesn't include phone_number claim
Hi guys! A little issue I found today, after feeling that someone was pulling my leg on the EspoCRM system;

## Summary

When `oidcSync` is enabled, every user login overwrites the `phoneNumber` field with `null` if the OIDC provider doesn't include the `phone_number` claim in its token. This silently wipes phone numbers that were manually set in EspoCRM.

## Steps to Reproduce

1. Enable OIDC authentication with `oidcSync` set to `true`
2. Use an OIDC provider that does **not** include `phone_number` in its claims (e.g., Cloudron)
3. Set a phone number on a User record in EspoCRM
4. Log out and log back in via OIDC
5. The phone number is gone

## Expected Behavior

Fields that are not present in the OIDC token should not be overwritten. If the OIDC provider doesn't include `phone_number`, the existing value in EspoCRM should be preserved.

## Actual Behavior

`Sync::getUserDataFromToken()` unconditionally returns `phoneNumber => null` (and `phoneNumberData => null`) when the claim is absent, which clears the existing phone number on every login.

## Root Cause

In `application/Espo/Core/Authentication/Oidc/UserProvider/Sync.php`, the method `getUserDataFromToken()` always includes all fields in its return array, even when the OIDC token doesn't contain the corresponding claims:

```php
private function getUserDataFromToken(UserInfo $userInfo): array
{
    return [
        'emailAddress' => $userInfo->get('email'),
        'phoneNumber' => $userInfo->get('phone_number'),   // returns null if claim missing
        'emailAddressData' => null,                         // always null
        'phoneNumberData' => null,                          // always null
        'firstName' => $userInfo->get('given_name'),
        'lastName' => $userInfo->get('family_name'),
        'middle_name' => $userInfo->get('middle_name'),
        'gender' => ...,
    ];
}
```

Then in `syncUser()`, this array is passed to `$user->set(...)`, which overwrites all these fields — including setting `phoneNumber` and `phoneNumberData` to `null`.

## Suggested Fix

Filter out null values before setting them on the user, so that missing OIDC claims don't overwrite existing EspoCRM data:

```php
private function getUserDataFromToken(UserInfo $userInfo): array
{
    $data = [];

    $map = [
        'emailAddress' => 'email',
        'phoneNumber' => 'phone_number',
        'firstName' => 'given_name',
        'lastName' => 'family_name',
        'middle_name' => 'middle_name',
    ];

    foreach ($map as $field => $claim) {
        $value = $userInfo->get($claim);
        if ($value !== null) {
            $data[$field] = $value;
        }
    }

    // Only set gender if it's a valid value
    $gender = $userInfo->get('gender');
    if (in_array($gender, ['male', 'female'])) {
        $data['gender'] = ucfirst($gender);
    }

    return $data;
}
```

This also affects `emailAddressData` and `phoneNumberData` which are unconditionally set to `null`, likely interfering with existing email/phone data structures.

## Environment

- EspoCRM 9.3.3
- OIDC provider: Cloudron (but any provider missing `phone_number` claim would trigger this)

## Workaround

Disable `oidcSync` in Administration > Authentication. This preserves OIDC login and auto-creation of new users (`oidcCreateUser`), but stops the profile overwrite on each login.

zigmasdirigeant:
Fixes #3638.

When `oidcSync` is enabled, `Sync::getUserDataFromToken()` returned every
field on every login, so any claim absent from the OIDC token (e.g.
`phone_number`, which Cloudron never sends) overwrote the existing EspoCRM
value with `null`. This silently wiped manually-entered phone numbers and
broke our Twilio VoIP routing, which reads the phone off the user profile.

Per the discussion on the issue, this distinguishes "claim absent" from
"claim present but null":

- `Payload::has()` and `UserInfo::has()` report claim presence. `Payload`
  stores all claims via `get_object_vars()`, so `array_key_exists()` is
  reliable and an explicit null still counts as present.
- `getUserDataFromToken()` only includes a field when its claim is present.

Resulting behaviour:

- Claim absent from the token -> field left untouched, existing value preserved.
- Claim present but null/empty -> field set to null, mirroring the provider's erase.

`emailAddressData` / `phoneNumberData` are now only reset when the
corresponding `email` / `phone_number` claim is present, instead of on
every sync.

CLA accepted in espocrm/cla#122.

### DIFF
--- a/application/Espo/Core/Authentication/Jwt/Token/Payload.php
+++ b/application/Espo/Core/Authentication/Jwt/Token/Payload.php
@@ -137,6 +137,15 @@ class Payload
         return $this->data[$name] ?? null;
     }
 
+    /**
+     * Whether a claim is present in the payload. Note that a claim explicitly
+     * set to null is considered present, as opposed to an absent claim.
+     */
+    public function has(string $name): bool
+    {
+        return array_key_exists($name, $this->data);
+    }
+
     public static function fromRaw(string $raw): self
     {
         $parsed = null;
--- a/application/Espo/Core/Authentication/Oidc/UserProvider/Sync.php
+++ b/application/Espo/Core/Authentication/Oidc/UserProvider/Sync.php
@@ -118,19 +118,39 @@ class Sync
      */
     private function getUserDataFromToken(UserInfo $userInfo): array
     {
-        return [
-            'emailAddress' => $userInfo->get('email'),
-            'phoneNumber' => $userInfo->get('phone_number'),
-            'emailAddressData' => null,
-            'phoneNumberData' => null,
-            'firstName' => $userInfo->get('given_name'),
-            'lastName' => $userInfo->get('family_name'),
-            'middle_name' => $userInfo->get('middle_name'),
-            'gender' =>
-                in_array($userInfo->get('gender'), ['male', 'female']) ?
-                    ucfirst($userInfo->get('gender') ?? '') :
-                    null,
-        ];
+        $data = [];
+
+        if ($userInfo->has('email')) {
+            $data['emailAddress'] = $userInfo->get('email');
+            $data['emailAddressData'] = null;
+        }
+
+        if ($userInfo->has('phone_number')) {
+            $data['phoneNumber'] = $userInfo->get('phone_number');
+            $data['phoneNumberData'] = null;
+        }
+
+        if ($userInfo->has('given_name')) {
+            $data['firstName'] = $userInfo->get('given_name');
+        }
+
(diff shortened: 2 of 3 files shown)


yurikuzn:
Thanks  for the PR.

We cannot accept significant changes like this PR introduces in the fix branch. 

With the master branch, these changes won't be compatible. Please see my last comment https://github.com/espocrm/espocrm/issues/3638#issuecomment-4602663524

Issue #3527, opened 2025-11-19
Logging in as another user gives Auth error when Espocrm is connected to Authelia


dacrudg:
**Describe the bug**
I'm using Docker EspoCRM 9.2.5-apache that is connected to authenticate through Authelia.  Logging in through Authelia works fine and as expected.  However, whenever you go to Administration -> Users -> choose a user then click the 3 dots and "Log In" as that user it will throw a red "Auth error".

**To Reproduce (MANDATORY, DO NOT REMOVE)**
Explicit steps to reproduce the behavior:
1. Connect EspoCRM to authenticate through Authelia
2. Login as admin using SSO / Authelia
3. Go to Administration -> Users -> choose a user then click the 3 dots and "Log in" as that user.  Once you go to login it will then throw red "Auth error"
...

**Expected behavior**
It should just login as the user you want

**Screenshots**

**EspoCRM version**
Docker EspoCRM 9.2.5-apache

Kharg:
Any errors in the logs?

dacrudg:
Only these:


```
1.2.3.4 - admin@domain.com [19/Nov/2025:19:55:56 +0000] "GET /api/v1/Oidc/authorizationData HTTP/1.1" 200 628 "https://espocrm.domain.com/?entryPoint=loginAs&anotherUser=user1@domain.com&username=admin@domain.com" "Mozilla/5.0 (X11; Linux x86_64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/142.0.0.0 Safari/537.36"

1.2.3.4 - - [19/Nov/2025:19:56:00 +0000] "GET /oauth-callback.php?code=authelia_ac_JWZpuadp1Kwt8Pla7bAtxohzS7mJMHzXcZtLca6Za-Y.KiL0um8s9DFQ1afL2x7gs4vsMqnF_Exdus85Ew0ZPvc&iss=https%3A%2F%2Fauthelia.domain.com&scope=openid+profile+email+groups&state=2s5v7t9y HTTP/1.1" 200 345 "https://authelia.domain.com/" "Mozilla/5.0 (X11; Linux x86_64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/142.0.0.0 Safari/537.36"

[...]

PR #3709, opened 2026-06-28
fix(impersonate): allow impersonate to work with oidc authentication

f-bor:
Fix for #3527


yurikuzn:
Thank you for the fix.

PR #3722, opened 2026-07-07
Fix numeric enum detail rendering

antonslyzhko:
Fixes incorrect detail-mode rendering for numeric enum fields without translations. Previously, **enumInt** and **enumFloat** values could fall through to the generic enum translation path. For values like **5**, this could display a character from the field name instead of the selected numeric value.

Example:
```json
"durationMinutes": {
    "type": "enumInt",
    "options": ["", 5, 30]
}
```

Selecting `5` could render `i` from `durationMinutes[5]` instead of `5`. The fix adds a shared numeric fallback translation map in **EnumIntFieldView**, which is also inherited by **EnumFloatFieldView**.

### DIFF
--- a/client/src/views/fields/enum-int.js
+++ b/client/src/views/fields/enum-int.js
@@ -39,6 +39,34 @@ class EnumIntFieldView extends EnumFieldView {
 
     validations = []
 
+    setup() {
+        super.setup();
+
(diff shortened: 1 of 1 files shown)


yurikuzn:
Good. I'll add 'private' to the method, as such methods should not be exposed.

PR #3731, opened 2026-07-17
Add theme direction parameter and improve RTL support

### LINKED ISSUE #3697: Split RTL direction from theme selection
**Is your feature request related to a problem? Please describe.**

Currently RTL support is tied to the dedicated `EspoRtl` theme. This means users who need an RTL interface, for example Arabic, Persian, Hebrew, or Urdu users, are effectively limited to one theme.

Direction is a language/layout concern, while theme is a visual preference. At the moment, these two concerns are coupled, so RTL users cannot use the other available themes in RTL mode.

**Describe the solution you'd like**

I would like RTL direction to be calculated from the selected language instead of requiring a separate RTL theme.

For RTL languages such as `ar_*`, `fa_*`, `he_*`, `ur_*`, EspoCRM could set a document/body marker, for example:

```html
<html dir="rtl">
```

or:

```html
<body data-direction="rtl">
```

Then common RTL CSS rules could be applied on top of the selected visual theme.

This would allow any theme to work in RTL mode, for example Light + RTL, Dark + RTL, Hazyblue + RTL, etc.

The separate `EspoRtl` theme could then be deprecated or kept temporarily for backward compatibility.

**Describe alternatives you've considered**

The current alternative is to continue using the `EspoRtl` theme. However, this limits RTL users to one theme and prevents combining RTL layout with other visual themes.

Another alternative is creating separate RTL variants for every theme, but this would duplicate theme maintenance and does not scale well.

**Additional context**

This change would make RTL support more flexible for multilingual installations. It would also separate two independent concerns:

- selected language determines text/layout direction;
- selected theme determines visual appearance.

This would improve the experience for Arabic and other RTL-language users without requiring a dedicated RTL-only theme.

If you are not available or does not have time to work on this, we would be happy to help implement it for EspoCRM. If this change is not preferred in the core, we can implement it as a custom module, but I believe core support would be the better practice because RTL direction is a general UI concern rather than an extension-specific feature.

eymen-elkum:
## TL;DR

This PR replaces the standalone `EspoRtl` theme with a `themeParams.direction` parameter (`ltr` or `rtl`). Direction is applied semantically through `<html dir>`, works with every Espo-derived theme, and is supported in Settings, Preferences, Calendar, Timeline, Lead Capture, and WYSIWYG iframes.

It also improves the previously partial RTL implementation, consolidates the required RTL Less into the base Espo stylesheet, adds logical pull utilities, and migrates existing `EspoRtl` configurations.

Related issue: #3697

## Main changes

- Added `themeParams.direction` with `ltr` as the default.
- Added direction selectors to Settings and Preferences.
- Added `<html dir="ltr|rtl">` and `body[data-direction]`.
- Removed the active `EspoRtl` theme and standalone RTL stylesheet structure.
- Added selector-gated RTL rules to the base Espo stylesheet.
- Added RTL support to Calendar, Timeline, Lead Capture, and WYSIWYG iframes.
- Added `.pull-start` and `.pull-end` logical utilities.
- Added a V10.1 migration and automated tests.

## Lead Capture direction

Lead Capture forms are standalone public pages without authenticated user Preferences. They expose a form language and visual theme, but no independent `themeParams.direction` setting.

Direction therefore follows the effective `formLanguage`, falling back to the system language:

- Arabic, Persian, Hebrew, and Urdu use RTL.
- Other languages use LTR.

This avoids adding another Lead Capture field and makes existing forms work without manual reconfiguration. The visual theme remains independent. Migrated `EspoRtl` forms use `Espo`, while their direction comes from their language.

## Logical pull utilities

Bootstrap's `.pull-left` and `.pull-right` are physical and do not adapt to document direction. Reversing them globally in RTL would be risky because some usages may intentionally refer to a physical side.

This PR adds:

```less
.pull-start {
    float: inline-start !important;
}

.pull-end {
    float: inline-end !important;
}
```

`.pull-end` means right in LTR and left in RTL.

Record detail and edit templates use `.pull-end` because `pull-right` was assigned directly in their markup. Fixing these cases only from RTL Less would require long, fragile contextual selectors and additional `!important` overrides.

I recommend gradually using logical pull utilities wherever alignment is semantic, while retaining physical pull utilities where a fixed side is intentional.

## Updating the bundled Bootstrap core

For small, generic utilities such as `.pull-start` and `.pull-end`, I recommend updating Espo's bundled Bootstrap Less directly.

Espo uses Bootstrap 3, whose upstream feature development has ended, so substantial future changes to this bundled version are not expected. A small backward-compatible addition is therefore less likely to conflict with upstream work than patching an actively developed library.

Keeping these utilities beside the existing physical pull utilities also avoids duplicating foundational behavior in `custom.less`. Existing Bootstrap behavior remains unchanged; larger or Espo-specific changes should continue to live in Espo's own Less files.

## Why `getVarParam` was updated

Previously, preference `themeParams` were read only when the user had explicitly selected a personal theme.

Direction is independent from the visual theme, so this valid configuration must work:

```text
theme: null
themeParams.direction: rtl
```

`getVarParam` now reads a preference parameter when the user either selected a theme or explicitly stored that parameter. It still respects `userThemesDisabled` and retains the existing configuration and metadata fallback order.

Without this change, a user inheriting the system theme could not have a personal direction.

## RTL Less organization

The old `espo-rtl` tree contained a broad Bootstrap RTL implementation and duplicated theme initialization. It has been replaced with a shared overlay:

```text
frontend/less/espo/rtl/
├── bootstrap.less
├── custom.less
├── layout-side.less
├── layout-top.less
└── main.less
```

All emitted rules are scoped under `html[dir="rtl"]`.

The old component files were reviewed individually. Required rules were retained and consolidated, while unused rules and behavior already handled by native RTL flow or current Espo styles were removed. For example, the unused `.pull-left.flip` and `.pull-right.flip` rules were dropped.

Some retained rules may become removable as more shared styles adopt logical properties. They were kept where aggressive removal could regress Bootstrap components or existing Espo layouts.

## CSS size

The target was to keep the minified RTL addition below 5% of the base stylesheet. The final result is:

- Base Espo CSS: `515,145` bytes
- RTL addition: `29,393` bytes
- RTL compared to base: `5.71%` (approximately 6%)

The target was exceeded slightly because this PR does more than split RTL into a parameter. The previous theme only partially supported RTL, so missing application-level corrections were also added.

Required layout rules were retained rather than removed only to meet the size target. They remain selector-gated and dormant in LTR mode.

## Additional RTL support

The PR extends RTL support to areas that were previously incomplete, including:

- Login, authenticated startup, and logout direction handling.
- FullCalendar using its native `direction` option.
- Timeline using its native `rtl` option.
- Calendar navigation controls and directional arrows.
- Lead Capture forms and WYSIWYG iframe documents.
- Side and top navbars, dropdowns, lists, panels, pagination, kanban, inline editing, and search controls.

Native Calendar and Timeline options are used for their internal layouts, with only small application-level corrections kept in Less.

## Migration

The V10.1 migration covers system configuration, user Preferences, portal users, Portal records, and Lead Captures.

For system, user, and portal settings:

- `EspoRtl` becomes `Espo`.
- `direction` becomes `rtl`.
- Existing `themeParams` are preserved.
- Explicit navbar values are preserved.
- A missing navbar becomes `top`, matching the old `EspoRtl` default.

For Lead Capture, `formTheme: EspoRtl` becomes `Espo`, and direction is derived from `formLanguage`.

Non-`EspoRtl` configurations are not modified.

## Testing

Coverage was added for direction resolution, invalid values, preference overrides without a personal theme, Lead Capture languages, and migration parameter preservation.

Local verification:

- PHP unit tests: `11 tests, 11 assertions` passed.
- Less compilation passed for all built-in themes.
- Theme CSS minification passed.
- RTL output remains scoped under `html[dir="rtl"]`.
- No standalone `EspoRtl` stylesheet is generated.


yurikuzn:
Thank you for the massive PR.

The first question that came to my mind after quickly scanning the code. It would be good if we used a PHP enum, like `...Utils\Theme\Direction::Rtl` instead of strings.

eymen-elkum:
Thanks, good idea. I’ve added `Espo\Core\Utils\Theme\Direction` as a backed enum and updated the PHP-side direction flow to use `Direction::Ltr` and `Direction::Rtl`. Strings remain only at configuration, metadata, persistence, and output boundaries.

Please see the latest commit.

yurikuzn:
I noticed an issue. In preferences, when the Default theme is selected, the parameter sub-fields should be hidden. Now, they are visible.

As it was before:

image


As it is now:

image

---

I pointed out what change caused this behavior the review comment above.

eymen-elkum:
Yes, you are right. I may have had to make this point as a suggestion and not push it to the PR.

The case is that users can change parameters while keeping the default theme. Even though, I believe this point is better addressed by you.

What do you suggest? Revert this change?

yurikuzn:
The original behavior was intentional, we should preserve it.

yurikuzn:
As well as the logic in theme-manager.js. [‎client/src/theme-manager.js‎](https://github.com/espocrm/espocrm/pull/3731/changes#diff-b5bceabcf21d6c0acbf70666c40afde238ce275a53d00376bd7c6d88ba217212).

We need to preserve the check.

```
 if (
    this.useConfig &&
    !this.config.get('userThemesDisabled') &&
    this.preferences.get('theme')
)
```

eymen-elkum:
Ready for review,

I noticed one issue on ui and I leave it for you as part of using `Logical pull utilities` so this PR review not being harder, see the screenshot.

Screenshot 2026-07-24 at 11 30 02 AM

Please let me know if you’re ok being a part of this PR.

yurikuzn:
This issue was present originally?

yurikuzn:
I don't observe this issue. Maybe it's a customization?

eymen-elkum:
Yes, attached is a fresh screenshot of official demo.

Screenshot 2026-07-24 at 11 42 48 AM

The correct must be:

Screenshot 2026-07-24 at 11 43 33 AM

yurikuzn:
This one is likely to be addressed with a CSS fix. Might be was not covered in the RTL's LESS before.

yurikuzn:
Everything else looks fine. The only thing is the lead capture's FormService using a separate method `getDirection`. I'll refactor so that it's returned in `getData` .

eymen-elkum:
Thanks for merging!

yurikuzn:
Thank you for the effort.

I refactored the form service: https://github.com/espocrm/espocrm/commit/1a3d2f1e0a84aa60d1e115e81b65c961a14c45f6

PR #3738, opened 2026-07-21
Fix navbar overflow width calculation

eymen-elkum:
## Summary

Fixes top-navbar layout issue where sometimes navigation items, including the **More** (`…`) menu, could wrap onto a second line and make the navbar approximately twice its intended height.

## Problem

`NavbarSiteView.adjustTop()` calculated the available width using the current width of the right-side navbar container.

During navbar adjustment, the browser’s flex layout could temporarily shrink that container. This produced an underestimated width for the right-side controls and incorrectly indicated that an additional navigation tab would fit.

Once the controls returned to their natural width, the remaining space was insufficient, causing:

- The **More** menu to wrap onto a second line.
- Right-side navbar items to wrap in some cases.
- The navbar height to increase from approximately 42 px to 82 px.
- Inconsistent behavior during initial rendering and window resizing.

## Root Cause

The right-side navbar width was:

- Calculated only once before repeated adjustment passes.
- Measured from a flex container that could be temporarily compressed.
- Not guaranteed to represent the natural combined width of its visible items.

As a result, `adjustTop()` could make its overflow decision using an unstable width.

## Solution

The navbar adjustment now:

- Recalculates the required right-side width during every adjustment pass.
- Measures each visible right-side navbar item using its natural outer width.
- Includes margins when calculating the total required width.
- Uses the stable combined width when determining how many tabs can remain visible.
- Moves additional tabs into the **More** menu when necessary.

This prevents temporary flex shrinking or late-rendered controls from affecting the available-width calculation.

## Screen Recording

https://github.com/user-attachments/assets/42a94eed-8dce-48ee-b566-fbf74b034a56

### DIFF
--- a/client/src/views/site/navbar.js
+++ b/client/src/views/site/navbar.js
@@ -648,6 +648,7 @@ class NavbarSiteView extends View {
         };
 
         const $navbar = $('#navbar .navbar');
+        const navbarRightElement = this.$navbarRight.get(0);
 
         if (window.innerWidth >= smallScreenWidth) {
             $tabs.children('li').each(() => {
@@ -657,8 +658,6 @@ class NavbarSiteView extends View {
             $navbar.css('max-height', 'unset');
             $navbar.css('overflow', 'visible');
         }
-
-        const navbarBaseWidth = this.navbarHeaderElement.clientWidth + this.$navbarRight.width();
 
         const tabCount = this.tabList.length;
 
@@ -672,6 +671,7 @@ class NavbarSiteView extends View {
         const updateWidth = () => {
             const windowWidth = window.innerWidth;
             const moreWidth = $moreLi.width();
+            const navbarRightWidth = navbarRightElement.offsetWidth;
 
             $more.children('li.not-in-more').each(() => {
                 unhideOneTab();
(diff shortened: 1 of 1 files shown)


yurikuzn:
Thanks for the PR.

If possible, could you rewrite it so that JQuery is not used, except for obtaining the element if needed. We have a policy of not using JQuery in new code.

yurikuzn:
Is necessary to iterate through children? Cannot we obtain the width of the container directly?

eymen-elkum:
You’re right; iterating through the children is not necessary. I’ve updated the implementation to obtain the container’s DOM element once and read its offsetWidth directly before restoring the hidden tabs. This also avoids using jQuery for the width calculation.
Going to squash 2 commits in one right now.

yurikuzn:
Thanks. Could you redirect it to the fix branch?

BTW I cannot reproduce. Does it requires some conditions?

eymen-elkum:
> BTW I cannot reproduce. Does it requires some conditions?

Attached above at end of first comment is a video demonstrating the issue. The screen recording was taken from the official demo site. The problem occurs infrequently when the remaining width for the tab list is very close to a group of tabs. In the past, I resolved this issue by adding gap in code. However, I believe this solution is more effective.

eymen-elkum:
Additionally, when this issue arises, the script continues to execute in intervals for recalculation, which also negatively impacts the overall performance.

eymen-elkum:
Rebased to `fix`.

PR #3749, opened 2026-08-07
fix(email): pre-select group from-address on reply

### LINKED ISSUE #1230: Reply email address
Hi,

I have multiple group email addresses (email boxes).
When I (or anyone) open a group email and reply to the email the from address always the personal email address so every time I have to update the from email to the group email.
How can I setup to have the reply email address actually the group email automatically?

Thanks
G

### LINKED ISSUE #2530: Email: Corresponding from-address pre-selected when replying on email - Group Email
If an email is sent to a group email address, replying on it will pre-select that address as an from-address. 

This is necessary for users who are part of different teams with different group email address.

meysam81:
replying to an email received on a group email account pre-selects the user's personal address as `From`.

currently the group address must be picked from the dropdown on **every** reply.

* fixes #2530                            
* fixes #1230                                              
* also fixes https://forum.espocrm.com/forum/developer-help/38492-limit-user-to-send-emails-from-specific-account-only

### DIFF
--- a/client/src/email-helper.js
+++ b/client/src/email-helper.js
@@ -97,6 +97,7 @@ class EmailHelper {
         const replyToAddressString = model.attributes.replyTo || null;
         const replyToString = model.attributes.replyToString || null;
         const userEmailAddressList = this.user.attributes.emailAddressList || [];
+        const lcUserEmailAddressList = userEmailAddressList.map(it => it.toLowerCase());
         const idHash = model.attributes.idHash || {};
         const typeHash = model.attributes.typeHash || {};
 
@@ -130,7 +131,7 @@ class EmailHelper {
             (!to || !to.includes('@')) &&
             model.attributes.from
         ) {
-            if (!userEmailAddressList.includes(model.attributes.from)) {
+            if (!lcUserEmailAddressList.includes(model.attributes.from.toLowerCase())) {
                 to = model.attributes.from;
 
                 if (!nameHash[to]) {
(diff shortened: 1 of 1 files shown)


yurikuzn:
Hi,

As it was mentioned in the issue, it was considered as not desired behavior for some use cases.

meysam81:
oh damn

sorry I was really frustrated to have to check which inbox I've already sent the email from

what's the blocker to get this in? if at all.

maybe we can find a fine balance to address both concerns somehow

yurikuzn:
For example, internally, we reply from personal accounts only. I assume it's the case for the most of organization. A customer expects to communicate with a person.

I'd appreciate it here we avoided emotionally charged language such as "damn" or "I was really frustrated." It's really tough to maintain all this, it makes developers hate what they are doing.

You can monkey patch the logic for now, so that it works as you need.

meysam81:
I was just expressing myself in my own term... no hard feelings

---

here's my use case just for clarity... I am the sole-tenant of my instance, having multiple accounts `john@example.com` and `bob@example.com`

every time I get a reply, it could be from either of the inboxes and I wanted to have that default when hitting `Reply`

anyway, if this is not a good fit, feel free to close it

best of wishes to you and thanks for your product and your hardwork

yurikuzn:
We need to introduce a parameter in the future, either system-level or user-level.

--- 
Or group-email-account level.

PR #3772, opened 2026-08-27
fix: prevent IME composition from erasing Int/Float/Currency field values

### LINKED ISSUE #2921: Int/Float/Currency input issue with IME enabled
When using a Chinese, Japanese or Korean keyboard language with IME enabled, after entering value and clicking outside of the input element, the entered value is erased. Happens on Windows 11 (works fine on 10), iPhone.

1. Use Windows 11 or iPhone, use Chinese, Japanese or Korean keyboard language with IME enabled.
2. Open the edit view of an Opportunity.
3. Enter a value in the Amount field.
4. Click outside of the input element.
5. The value is erased.

**Expected behavior**
The value is not erased.

**EspoCRM version**
8.0, the *master* branch.

**Additional context**
We use **AutoNumeric** library for number inputs.

evan70:
## Problem

When using CJK (Chinese, Japanese, Korean) keyboards with IME enabled, clicking outside an Int/Float/Currency input during composition erases the entered value. This affects Windows 11 and iPhone users.

## Root Cause

AutoNumeric processes input events during IME composition, before the composition is finalized. When the user clicks outside the input while IME is composing, the partially-composed value is read and processed, causing the final value to be erased.

## Fix

Track / events on the input element:

1. Set  on 
2. Set  on  and re-sync AutoNumeric with the final composed value via 

Applied to both edit mode and search mode.

## Changes

- `client/src/views/fields/int.ts` — Added IME composition event listeners and `isComposing` state tracking

## Testing

- Verified IME composition works correctly on Chrome/Edge with Chinese input
- Verified non-IME input still works as before
- Verified search mode fields are also fixed

Closes #2921

### DIFF
--- a/client/src/views/fields/int.ts
+++ b/client/src/views/fields/int.ts
@@ -101,6 +101,8 @@ class IntFieldView<
 
     protected disableFormatting: boolean = false
 
+    protected isComposing: boolean = false
+
     protected setup() {
         if (this.getPreferences().has('thousandSeparator')) {
             this.thousandSeparator = this.getPreferences().get('thousandSeparator');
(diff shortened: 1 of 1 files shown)


yurikuzn:
Hi,

Thanks for PR. It seems that the issue is still reproducible after the fix.

PR #3790, opened 2026-09-15
Bool filters onlyMy and shared: use relation conditions in middle join

Dev-next-gen:
While reading fc7ac8bd1b ("Fix only team filter, use conditions") I checked the other places that join the assignedUsers and collaborators middle tables, and found two that were left out: the `onlyMy` and `shared` bool filters.

Both relations are stored in the shared `EntityUser` and `EntityCollaborator` tables, and their relation defs carry an `entityType` condition (set in `LinkConverters/EntityUser.php` and `EntityCollaborator.php`). `OnlyMy` and `Shared` build the join with only `entityId` and `deleted`, so the sub-query matches a middle row by id alone, whatever entity type it belongs to. With the usual random ids this rarely shows up, but it can give a false match when two records of different types share an id, and it is inconsistent with `OnlyTeam`, `ForeignOnlyOwn`, `ForeignOnlyTeam` and `RelationQueryHelper`, which all constrain `entityType`.

The change copies the loop from `OnlyTeam`: the relation conditions are added to the join conditions, so nothing changes for relations that have none.

I added `tests/unit/Espo/Core/Select/Bool/FiltersTest.php`, which builds each filter against a many-many relation with an `entityType` condition and checks the join. Without the fix both tests fail (`Failed asserting that null matches expected 'Test'`), with it they pass, and the whole `tests/unit/Espo/Core/Select` suite passes (135 tests). phpstan reports no errors on the two changed files. I based this on `fix` since it is a small fix; happy to retarget it to `master` if you prefer.

Found by a defect-hunting pipeline I build and run ([Dev-next-gen](https://github.com/Dev-next-gen)), using Claude Code with Anthropic's Claude Opus 5.


yurikuzn:
Hi,

Thank for the PR. Could you accept our CLA by creating a commit here: https://github.com/espocrm/cla.

Dev-next-gen:
Done — signed in espocrm/cla#130, adding `contributors/Dev-next-gen.md` as your README describes.

Thanks for looking at the PR so quickly.

